PR #6, opened 2017-09-07
libris/lxltools#2 - Fix error with ZipFile-PosixPath incompatibility

mblomdahl:
Fixes libris/lxltools#2 (preventing me from getting LibrisXL ES/PSQL in shape).

### DIFF
--- a/datasets.py
+++ b/datasets.py
@@ -34,7 +34,7 @@ def to_camel_case(label):
 
 
 def _get_zipped_graph(path, name):
-    with zipfile.ZipFile(path, 'r') as zipped:
+    with zipfile.ZipFile(str(path), 'r') as zipped:
         return Graph().parse(zipped.open(name), format='turtle')
 
 


mxtthias:
Can you add some info in the commit message about why this is needed? E.g. offending package versions, etc.

mblomdahl:
Explanation:

> After lxltools.datacompiler.Compiler changed its behavior to returning "path" as a pathlib PosixPath, ZipFile(path) stopped working due to https://bugs.python.org/issue28231 (only resolved in Python 3.5+, apparently)

Would this suffice?

mxtthias:
Explanation looks good.

Does this work (without the fix) if you run it with Python 2.7?

mblomdahl:
@mxtthias No, that's where the problem occurs – in Python 2.7; locally, on my CentOS 7 box and @jannistsiroyannis's computer as well.

mxtthias:
Just so I understand this, why did the issue first appear now? Something about existing dependencies not being updated?

mblomdahl:
> Something about existing dependencies not being updated?

I think so. Jannis were trying out the `fab conf... app.whelk.import_definitions` on the server and his computer, everything worked. Then he cleared the old directories that was (since long) checked out and ran the procedure again, ending up with the same libris/lxltools#2 error. @jannistsiroyannis Can you corroborate?

jannistsiroyannis:
I can confirm running into the same problem with a freshly checked out repository and virtualenv (but my old cached version still works).

mxtthias:
Check.

PR #9, opened 2018-01-05
Add chip definition for Relationship type

oBlissing:
Show property "entity"

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -134,6 +134,11 @@
           "@type": "fresnel:Lens",
           "classLensDomain": "Contribution",
           "showProperties": [ "agent", "role" ]
+        },
+        "Relationship": {
+          "@type": "fresnel:Lens",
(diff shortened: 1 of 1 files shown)


source/vocab/display.jsonld
```diff
@@ -134,6 +134,11 @@
           "@type": "fresnel:Lens",
           "classLensDomain": "Contribution",
           "showProperties": [ "agent", "role" ]
+        },
+        "Relationship": {
+          "@type": "fresnel:Lens",
+          "classLensDomain": "Relationship",
+          "showProperties": [ "entity" ]
```
niklasl: We should add `"relation"` to `"showProperties"` too (it's related to `role` but used for document relationships).

oBlissing: I was thinking about this, but couldn't wrap my head around how that would look. Maybe we should just add that one too and see :)

niklasl: Go for it! :)

PR #10, opened 2018-05-07
Update chips of Organisation, Jurisdiction and Meeting

sarwik:
Add marc:subOrdinateUnit and isPartOf { name }

It's missing supports of the model when marc:subOrdinateUnit exists and therefore only displays "{Namnlös entitet}"

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -109,7 +109,7 @@
         "Jurisdiction": {
           "@type": "fresnel:Lens",
           "classLensDomain": "Jurisdiction",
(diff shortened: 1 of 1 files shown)


niklasl:
The isPartOf: name structure should not work. Does lxlviewer support that? Just "isPartOf" would be the correct rule. (The value of that list is of `@type: @vocab` in JSON-LD.)

(Also, didn't we remove subordinateUnit?)

klngwll:
niklas du glömde

2018-05-07 15:56 GMT+02:00 Niklas Lindström <notifications@github.com>:

sarwik:
Ok, I will change it. 

(No, we didn't remove subordinateUnit, because it created new revert issues, which there was no time to solve before release)

PR #11, opened 2018-05-08
Make schemes notation lower-case for 'barn', since it is used as code

sarwik:
Since notation is used as code, it needs to be lower-case to be able to match (for instance at revert).


sarwik:
It is supposed to be merged to the release branch, not develop. Closing this pull request....

sarwik:
I noticed barngf... but a bit to late ;) But I have changed it now. Good that you noticed it!
I create a jira-issue for gmgpc-swe so we can verify it.

PR #16, opened 2018-05-16
Feature/update electronic locator

sarwik:


source/marc/terms.ttl
```diff
@@ -579,9 +583,22 @@ marc:city a owl:DatatypeProperty ;
...
+    sdo:domainIncludes kbv:Document, kbv:MediaObject, kbv:Electronic .
+
+marc:electronicName a owl:DatatypeProperty ;
+    rdfs:label "Elektroniskt namn"@sv ;
+    sdo:domainIncludes kbv:Document, kbv:MediaObject, kbv:Electronic .
+
 marc:linkText a owl:DatatypeProperty ;
     rdfs:label "Länktext"@sv ;
-    rdfs:domain kbv:Note .
+    rdfs:domain kbv:Note, kbv:Document, kbv:MediaObject, kbv:Electronic .
```
niklasl: This needs to be domainIncludes (otherwise the domain is implied to be *all* of those).

source/marc/terms.ttl
```diff
@@ -26,6 +26,7 @@ marc:relatedAs a owl:ObjectProperty ;
     rdfs:range kbv:ObjectProperty .
 
 marc:electronicLocatorType a owl:DatatypeProperty ;
+    sdo:domainIncludes kbv:Document, kbv:MediaObject, kbv:Electronic ;
```
niklasl: If you add: `:MediaObject rdfs:subClassOf :Document` where `:MediaObject` is declared (relations.ttl AFAIK) and verify that :Electronic is a subclass of document (via :Instance), we can use just `rdfs:range kbv:Document` throughout in this change.

source/marc/terms.ttl
sarwik: What a blunder. I'll fix.

source/marc/terms.ttl
sarwik: Good idea. I'll fix.

PR #17, opened 2018-05-17
Add 780-785 relations missing from BF2

klngwll:
Links to BF1 vocabulary (bibframe.org) is not working at the moment, don't know if it is temporary error or if LC has taken it down. I have therefore not created any equivalent statements.


klngwll:
Also testing assignee vs reviewer :)

PR #19, opened 2018-05-17
Feature/add rda terms

niklasl:
Add RDA terms with mappings to LoC and RDA Registry.


niklasl:
There's an AudioInstance too. :)

PR #51, opened 2018-07-04
Remove marc enumerations that should not be used anymore

sarwik:
Applies to marc enumerations that have been removed from marcframe (when marcDefault has been added). The enumerations are therefore no longer mapped in marcframe and can not be reverted.

The enumerations were previously mapped to:
* auth 710/711/730/748: marc:SourceNotSpecified
* auth 377: marc:ISO-639-2B
* bib 033: marc:NoDateInformation


klngwll:
We need to check revert behavior on existing data with these after merge into develop.

sarwik:
True! I will fix an issue for it in jira so we won't miss it

PR #54, opened 2018-07-09
Update range for termComponentList to Identity

klngwll:
Previous range with only subject and subdivision was too narrow. Need to be able to handle Works, Agents, Temporals etc. Identity covers all the cases but might also be to broad. Need user input if further tweaks should be done.


niklasl:
Do we want to do some testing prior to merging, or back out of this if reactions are negative? (The alternate solution is to cherry-pick desired classes to rangeIncludes.)

PR #65, opened 2018-10-24
Make descriptionUpgrader a set

sarwik:


sarwik:
Align with https://github.com/libris/librisxl/pull/245

klngwll:
How will this affect adding more than one descriptionUpgrader to existing data?

sarwik:
Existing data only consists of one entity instead of a list of entities. New data will instead consist of a list.
I have tested the behavior (in viewer) of adding a new entity to descriptionUpgrader to existing (imported) data (which only consist of one entity). It works fine; it is possible to add a new entity and the format will instead become a list.
Exporting existing data also works fine.
So if we, for now, are ok with that imported data from voyager and new data not will have the exact same format (no list vs list) I can not see any problems with changing marcframe from "link" to "addLink" and making descriptionUpgrader repeatable in viewer. However, we should write a jira issue for the difference in format of old and new data.
It is also important to test, more carefully in Dev and QA, the export of existing data and to add new entity in viewer to descriptionUpgrader.

PR #73, opened 2018-11-19
Feature/cartographic still image updates

sarwik:


sarwik:
I have tested it locally (together with the templates updates) and it looks good

PR #76, opened 2018-12-11
Feature/general definition cleanups

klngwll:
- Removed english labels that were "almost the same" in BF2 but became doubled in english version of the interface.
- Fixed faulty BF2 mapping of MatrixNumber.
- MusicEnsemble now has correct bf2 prefix
- ISBN13 is now pending so it isn't addable through the interface. Not used in conversion so might actually be an ancient artefact.
- Language, now doesn't include identifier in it's range :)

### DIFF
--- a/source/vocab/agents.ttl
+++ b/source/vocab/agents.ttl
@@ -277,8 +277,6 @@
     owl:equivalentClass sdo:Library .
 
 :bibliography a owl:ObjectProperty ;
-    :category :compositional ;
-    # until we can properly link bibliographies, when we do we also probably want to restrict the creation of new ones.
     rdfs:domain :Record ;
     rdfs:range :Library ;
     rdfs:label "bibliografi"@sv .
--- a/source/vocab/agents.ttl
+++ b/source/vocab/agents.ttl
@@ -277,6 +277,8 @@
     owl:equivalentClass sdo:Library .
 
 :bibliography a owl:ObjectProperty ;
+    :category :compositional ;
+    # until we can properly link bibliographies, when we do we also probably want to restrict the creation of new ones.
     rdfs:domain :Record ;
     rdfs:range :Library ;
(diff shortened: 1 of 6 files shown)


source/vocab/relations.ttl
```diff
@@ -404,10 +404,10 @@
     owl:equivalentProperty foaf:isPrimaryTopicOf .
 
 :language a owl:ObjectProperty;
-    rdfs:label "language"@en, "språk"@sv;
+    rdfs:label "språk"@sv;
     # TODO worklanguage? 
     rdfs:comment "Språk förknippat med en resurs eller dess delar."@sv;
-    sdo:domainIncludes :Creation, :StructuredValue;
+    sdo:domainIncludes :Creation, :Summary, :TableOfContents, :Title, marc:Libretto, marc:SubtitlesOrCaptions, marc:SungOrSpokenText;
```
niklasl: I'm wondering if we want an (abstract) subclass of StructuredValue as a base class for some of these? Possibly a base for Note as well, which reasonably can have a language. I'd call it `Note`, but that's obviously taken... Perhaps `Annotation`, to bring back an old BF1-term...

klngwll: I think we first should figure out what the Marc classes actually are about and their place in the model, then I need to remove Title which already is a StructuredValue :)

klngwll: Never mind that brainmelt, I removed SV to fix this :)

source/vocab/agents.ttl
```diff
@@ -277,6 +277,8 @@
     owl:equivalentClass sdo:Library .
 
 :bibliography a owl:ObjectProperty ;
+    :category :compositional ;
```
niklasl: This is not logically compositional. We should not "hack" the vocab for interface needs like this (it will just shadow the lack of linkable bibliographies).

PR #80, opened 2019-01-23
Remove bibLevel, BibLevelType and typeOfRecord from marc legacy terms

niklasl:


niklasl:
16 occurrences of marc:bibLevel in existing records, 0 of marc:typeOfRecord. I'm writing off the 16 as noise (all but 2 link to https://id.kb.se/marc/MonographItem, which we imply for issuanceType if missing), but report them informally to the metadata-team for possible cleanup.

PR #81, opened 2019-01-23
Feature/display maintenance

klngwll:
Update display file for view and indexing.

* Add musicMedium to Work card so this property which is part of a title is searchable and correctly placed within work properties.

Add isPartOf to Work card to so locally described Works with this property is searchable.

* Add hasVariant to Person/Organization/Meeting/Jurisdiction cards so alternative names for agent is searchable.

* Add seriesmembership to Instance card for searchability of series information.


niklasl:
I trust you!

PR #83, opened 2019-01-28
Change label of code 'pro' to 'Producent'

sarwik:


sarwik:
The discussion began in the absence of a code for the music producer. You have a good point whether pro have been used instead of fmp. I have created a jira ticket for this (LXL-2317) so we can follow it up.

PR #84, opened 2019-02-05
Add enums that are mandatory for import to issn

sarwik:


sarwik:
Good point. I have verified that "_" actually means " ". I have followed the same pattern as encodingLevel which in vocab declares " "  as "_". However, I have not been able to test it locally. Regarding marcframe, I can't see there is anything we need to update since there is no tokenmap for this enumeration.

niklasl:
Encoding-level is not picked from the enums since it is defined in marcframe. Why can't this be tested locally?

source/marc/construct-enums.rq
```diff
@@ -1242,6 +1242,9 @@ construct {
         (marc:SerialsFrequencyType	marc:BimonthlyIncludes67Or8NumbersAYear	marc:SerialsFrequencyType-b	"b"	UNDEF	UNDEF	"Varannan månad (6-8 ggr/år)"@sv	"Bimonthly Includes 6, 7, or 8 numbers a year."@en)
         (marc:SerialsFrequencyType	marc:ThreeTimesAWeek	marc:SerialsFrequencyType-i	"i"	UNDEF	UNDEF	"Tre gånger i veckan"@sv	"Three times a week"@en)
         (marc:SerialsFrequencyType	marc:MonthlyIncludes91011Or12NumbersAYear	marc:SerialsFrequencyType-m	"m"	UNDEF	UNDEF	"Varje månad (9-12 ggr/år)"@sv	"Monthly Includes 9, 10, 11, or 12 numbers a year."@en)
+        (marc:SerialsFrequencyType	marc:NoDeterminableFrequency	marc:SerialsFrequencyType-_	"_"	UNDEF	UNDEF	"Frekvens kan ej fastställas"@sv	"No determinable frequency, Frequency is intentionally irregular."@en)
```
niklasl: Maybe shorten the English label too (i.e. remove "is intentionally irregular", since it's an implied "or" there and thus of no definitive meaning)?

sarwik:
After discussing with ISSN experts, we agreed to only add marc:UnknownFrequency ('u') and marc:OtherFrequency ('z'). I have tested it locally and it works as expected.

source/marc/construct-enums.rq
klngwll: It could be noted in a skos:comment something like "Used when the frequency is known to be intentionally irregular."

klngwll: NEVERMIND :)

niklasl:
The 'u' will be represented as the *absence* of a `frequency` , so we don't need an enum for that.

sarwik:
@niklasl Ah that's true. I update it.

PR #90, opened 2019-02-19
marc:SerialsRegularityType no longer subClass of Frequency

sarwik:
This disables the possibility to code frequency with
enums of type marc:SerialRegularityType


sarwik:
Tested locally together with https://github.com/libris/librisxl/pull/304 and it works as expected: marc:SerialsRegularityType is no longer subClass of Frequency and therefore not possible to add to frequency (for serials). Default value of export is set to '|'.

PR #97, opened 2019-03-05
Feature/add marc as scheme

sarwik:
* Add marc as scheme
* Add inScheme marc to genreForm enums
* Add exactMatch to LC's marcgt for genreForm enums where it is possible


source/schemes.ttl
```diff
@@ -232,6 +233,10 @@
     dc:title "Library of Congress Subject Headings"@en;
     foaf:homepage <http://id.loc.gov/authorities/subjects/> .
 
+<marc> a skos:ConceptScheme;
+    skos:notation "marc";
+    dc:title "MARC Genre Terms List"@en, "MARC Genre/Form Termlista"@sv .
```
niklasl: The `</term/marc>` scheme can't be just genre/form terms. These are all terms under `</term/marc>`.

source/marc/construct-enums.rq
```diff
@@ -62,291 +63,293 @@ construct {
         owl:deprecated ?deprecated ;
         skos:notation ?code ;
         skos:prefLabel ?label_sv, ?label_en ;
+        skos:inScheme ?inScheme ;
```
niklasl: I think we can get by with just:
```
        skos:inScheme :marc ;
```

source/schemes.ttl
niklasl: Actually, `</marc>`. For all "MARC terms". One pile.

PR #99, opened 2019-03-06
Feature/vocab update

klngwll:
Some vocab updates:

* Remove Geographic equivalence to Place since it is already a subClass.
* Update comments on nonfilingChars and hasVariant.
* Add marc:ComputerFileCharacteristics with a label.
* Add english labels for Nationality and Country.
* Indentation cleanups.
* Add pending to creditsNote since it is not used.

### DIFF
--- a/source/marc/terms.ttl
+++ b/source/marc/terms.ttl
@@ -355,7 +355,10 @@ marc:publicNote a owl:DatatypeProperty ;
                        marc:TextualHoldingsBasicBibliographicUnit,
                        kbv:Document .
 
-
+# BIB 256
+marc:ComputerFileCharacteristics a owl:Class ;
+    rdfs:label "Computer file characteristics"@en, "Filspecifika uppgifter"@sv ;
+    skos:closeMatch kbv:Note .
 # BIB 300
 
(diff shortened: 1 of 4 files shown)


source/marc/terms.ttl
```diff
@@ -355,7 +355,10 @@ marc:publicNote a owl:DatatypeProperty ;
                        marc:TextualHoldingsBasicBibliographicUnit,
                        kbv:Document .
 
-
+# BIB 256
+marc:ComputerFileCharacteristics a owl:Class ;
```
niklasl: I'd prefer the pattern we use for marc notes, i.e. ending with `Note`, but it isn't crucial (as in I don't expect this symbol to take on another meaning, separate from a kind of note).

source/vocab/concepts.ttl
```diff
@@ -17,10 +17,10 @@
 
 :Place a owl:Class;
     rdfs:label "Place"@en, "Plats"@sv;
-    owl:equivalentClass bf2:Place, prov:Location, sdo:Place, madsrdf:Geographic .
+    owl:equivalentClass bf2:Place, prov:Location, sdo:Place .
```
niklasl: :+1:

PR #103, opened 2019-04-02
Feature/separate classification topic schemes

sarwik:
* Add ClassificationScheme, GenreFormScheme and TopicScheme as subclasses of ConceptScheme
* Add restrictions to inScheme for types Classification, GenreForm and Subject. Use someValuesFrom 
  (instead of allValuesFrom) to not exclude the possibility to add schemes of type ConceptScheme 
* Change labels of inScheme and ConceptScheme
* Update schemes to subclass where it is applicable
* Add kao//eng as TopicScheme
* Remove scheme 'fao' in favour of 'agrissc'. (NB. There is no existing bib/hold records with link to 'fao' 
   and only few, 6 records, with local entity with prefLabel='fao' and inScheme='agrovoc')

Tested locally:
* Link to to scheme with restriction
* Link to new schemes: 'agrissc' and 'kao//eng'
* Revert ("Förhandsgranska Marc")

This PR needs to be aligned with PR in lxlviewer of corresponding updates of snippets and templates.

### DIFF
--- a/source/schemes.ttl
+++ b/source/schemes.ttl
@@ -217,6 +217,10 @@
 
 #@base <https://id.kb.se/ext/term/> . # ?
 
+<agrissc> a :ClassificationScheme;
+    skos:notation "agrissc";
+    foaf:homepage <http://http://agris.fao.org/agris-search/index.do> .
+
 <agrovoc> a :TopicScheme;
     dc:title "AgroVoc"@en;
     skos:notation "agrovoc" .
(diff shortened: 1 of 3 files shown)


klngwll:
Should we add exactMatch for schemes we define to the id.loc.gov representation like http://id.loc.gov/vocabulary/subjectSchemes/kao etc ?

klngwll:
Sorry, bad example since these are for http://id.loc.gov/vocabulary/classSchemes.html

sarwik:
Wohoo! 🎈

PR #115, opened 2019-05-08
Feature/update enums display

sarwik:
Only display prefLabel for chip marc:EnumeratedTerm.
For more info see LXL-2515 and LXL-2491.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -68,13 +68,13 @@
           "@id": "MediaType-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "MediaType",
-          "showProperties": [ "prefLabel" ]
+          "showProperties": [ "prefLabel", "label" ]
         },
         "ContentType": {
           "@id": "ContentType-chips",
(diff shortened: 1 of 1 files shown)


sarwik:
@oBlissing Check. I was not sure about the marc-prefix, if it would break something when the @id then would contain a `:` (colon). Since the UI only uses ids for extensions and there are none for marc:EnumeratedTerm, I merge it as it is now, but will double check it.

source/vocab/display.jsonld
```diff
@@ -231,6 +231,12 @@
           "@type": "fresnel:Lens",
           "classLensDomain": "TableOfContents",
           "showProperties": [ "label" ]
+        },
+        "marc:EnumeratedTerm": {
+          "@id": "EnumeratedTerm-chips",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "EnumeratedTerm",
```
niklasl: classLensDomain here is `marc:EnumeratedTerm`. (The `@id` is local and opaque, and is fine as it is.)

PR #119, opened 2019-05-31
Remove obsolete or OCLC specific encodingLevels

sarwik:
* Align this PR with https://github.com/libris/librisxl/pull/352
* See LXL-2545 and LXL-2546 for more info


klngwll:
Also, was there a reason to keep marc:EncLevelType-e?

sarwik:
@klngwll and @niklasl I wasn't sure when the conditions are marc legacy and might become noise. But definitely agree that by comment them out with an explanation, it will be easier to trace in the future.

sarwik:
> Also, was there a reason to keep marc:EncLevelType-e?

Well spotted! I missed it because we don't have any Instance with this enum ;)

PR #120, opened 2019-06-03
Add more information to encodingLevels

klngwll:
* Add additional information to meaning and usage of Classes through notes and comments.
* Fine tune swedish labels

### DIFF
--- a/source/marc/enums.ttl
+++ b/source/marc/enums.ttl
@@ -892,7 +892,7 @@
 
 :HoldingsLevel2 a :HoldingEncLevelType ;
     #owl:sameAs :HoldingEncLevelType-2 ;
-    :category :pending ;
+    kbv:category kbv:pending ;
     rdfs:comment "Används t.v. ej"@sv;
     skos:notation "2" ;
     skos:prefLabel "Holdings level 2"@en, "Beståndsnivå 2"@sv .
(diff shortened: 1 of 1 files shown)


source/marc/enums.ttl
```diff
@@ -812,58 +812,63 @@
...
     skos:prefLabel "Minimal level"@en, "Biblioteksnivå"@sv .
 
 :PrepublicationLevel a :EncLevelType ;
     #owl:sameAs :EncLevelType-8 ;
-    rdfs:comment "Post upprättad innan objektet givits ut"@sv;
+    rdfs:comment "Post upprättad innan resursen givits ut. Hit även CIP-poster."@sv;
     skos:relatedMatch <http://id.loc.gov/vocabulary/menclvl/8>;
     skos:notation "8" ;
-    skos:prefLabel "Prepublication level"@en, "CIP-post"@sv .
+    skos:prefLabel "Prepublication level"@en, "Förhandsinformation"@sv .
```
sarwik: Have everyone agreed to change prefLabel to "Förhandsinformation", even if we don't supplement with hovering info (that it also applies to CIP records)?

klngwll: No feedback so far, let's merge so it is visible in our dev environments for evalutation.

PR #124, opened 2019-06-25
Extend restrictions on instanceOf for MovingImageInstance/VideoRecording

sarwik:


sarwik:
I close this PR in favour of changing mapping of ProjectedImage (and all existing occurrences of it) to MovingImage. See LXL-2601 and LXL-2602 for more info.

PR #147, opened 2019-10-14
Feature/projected to moving image

klngwll:
* Remove ProjectedIMage references from Vocab.
* Make Visual class abstract.

Dependent on https://github.com/libris/lxlviewer/pull/507, https://github.com/libris/librisxl/pull/416

### DIFF
--- a/source/vocab/things.ttl
+++ b/source/vocab/things.ttl
@@ -238,7 +238,7 @@
 
 :Visual a owl:Class;
     owl:sameAs :Image ;
-    #ptg:abstract true ;
+    ptg:abstract true ;
     rdfs:label "Visual"@en, "Visuell resurs"@sv ;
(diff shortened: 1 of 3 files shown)


source/vocab/things.ttl
```diff
@@ -524,8 +523,8 @@
     rdfs:label "Projected Image"@en, "Projicerad bild"@sv;
     skos:exactMatch rdamedia:1005 ; # "projected" ("projected graphic", "motion picture")
     :inCollection marc:typeOfRecord;
-    # TODO: Work should not be necessary to set as subClassOf since Visual is subClass of Work. Should be removed with LXL-2735
```
niklasl: For a clean revert, we may want to remove the workaround in a separate commit after LXL-2735 has landed. @sarwik any opinions?

PR #149, opened 2019-10-21
Feature/fix languages

niklasl:


olovy:
We leave labels contaning pipe characters as is. Should they be converted to lists?
```
<http://id.loc.gov/vocabulary/iso639-2> <http://www.w3.org/2004/02/skos/core#hasTopConcept> <http://id.loc.gov/vocabulary/iso639-2/nno> . 
<http://id.loc.gov/vocabulary/iso639-2/nno> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2004/02/skos/core#Concept> . 
<http://id.loc.gov/vocabulary/iso639-2/nno> <http://www.w3.org/2004/02/skos/core#prefLabel> "Norwegian Nynorsk | Nynorsk, Norwegian"@en .
```
```
      "prefLabelByLang": {
        "de": "Nynorsk",
        "en": "Norwegian Nynorsk | Nynorsk, Norwegian",
        "fr": "norvégien nynorsk | nynorsk, norvégien",
        "sv": "Nynorska"
      }
```

source/construct-languages.rq
```diff
@@ -22,17 +22,20 @@ construct {
...
         } union {
             ?s skos:notation ?not .
             bind(strdt(?not, dc:ISO639-2) as ?langCode)
             bind(iri(concat('http://id.loc.gov/vocabulary/iso639-2/', ?not)) as ?loclang)
             graph <http://id.loc.gov/vocabulary/languages> {
-                ?loclang skos:prefLabel ?prefLabel .
+                ?loclang skos:prefLabel ?prefLabels
+                values ?n { 0 1 2 3 4 5 }
+                bind(concat("^([^|]+\\|){", str(?n) ,"} *") as ?skipN)
+                bind(replace(replace(?prefLabels, ?skipN, ""), " *\\|.*$", "") as ?prefLabel)
```
olovy: This works but is super slow (1m20s on my laptop)
Should be possible to make it conditional on `contains(?prefLabels, "|")`??

source/construct-languages.rq
```diff
@@ -3,41 +3,58 @@ prefix xsd: <http://www.w3.org/2001/XMLSchema#>
...
+                    filter(!contains(?prefLabel, "|"))
+                    ?loclang skos:prefLabel ?prefLabel
+                }
+                optional {
+                    filter(contains(?prefLabels, "|"))
+                    ?loclang skos:prefLabel ?prefLabels
+                    values ?n { 0 1 2 3 4 5 }
+                    bind(concat("^([^|]+\\|){", str(?n) ,"} *") as ?skipN)
+                    bind(replace(replace(?prefLabels, ?skipN, ""), " *\\|.*$", "") as ?prefLabel)
+                }
```
olovy: `var apf:strSplit (arg arg)`
http://jena.apache.org/documentation/query/library-propfunc.html

:sob:

niklasl:
@olovy we have a domain problem in that `prefLabel`, [according to SKOS](https://www.w3.org/TR/skos-reference/#L2831) is not allowed more than one value per language. So we need to move the one after the first `|` (are there ever more than two?) into an `altLabel`.

olovy:
> You can deregexpify this using `strbefore` and `strafter` (https://www.w3.org/TR/sparql11-query/#func-strbefore).

What about whitespace around `|`?

PR #152, opened 2019-11-19
Feature/sub class of from bf2

sarwik:
Cleanups of subClassOf values:
* If a class has more than one base class declared with subClassOf; keep only the most specific base class. This has occurred when bf2:subClassOf values have been included to our vocabulary and we have a more specific base class. See examples StillImage and MovingImage.
* Some minor corrections of misspelled 'subClassOf'


klngwll:
Excellent!

PR #153, opened 2019-11-19
Feature/display update

klngwll:
Updates to Display:

* Update Publication Chip with missing values for Serials and add Card.
* Add GenreForm and IntendedAudience to Instance Card.
* Add IntendedAudience, ContentType and IllustrativeContent to Works Card.
* Add KeyTitle and marc:SystemDetailsNote, marc:LanguageNote and marc:AddedHierarchicalPlaceName Chips for presentation.
* Update Identifier and ISBN Cards and Chips.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -413,6 +413,13 @@
           "classLensDomain": "CarrierType",
           "showProperties": [ "prefLabel", "label", "code", "termGroup" ]
         },
+        "Publication": {
+          "@id": "Publication-cards",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "Publication",
+          "fresnel:extends": {"@id": "Publication-chips"},
+          "showProperties": [ "fresnel:super", "marc:sequenceStatus", "hasPart" ]
+        },
         "ContentType": {
           "@id": "ContentType-cards",
           "@type": "fresnel:Lens",
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -136,7 +136,7 @@
           "@id": "Publication-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "Publication",
-          "showProperties": [ "place", "agent", "year", "date" ]
+          "showProperties": [ "place", "agent", "year", "startYear", "endYear", "date", "marc:publicationStatus" ]
         },
         "Serial": {
           "@id": "Serial-chips",
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -126,6 +126,12 @@
           "classLensDomain": "Identifier",
           "showProperties": [ "value", "typeNote", "marc:hiddenValue" ]
         },
+        "ISBN": {
+          "@id": "ISBN-chips",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "ISBN",
+          "showProperties": [ "value", "qualifier" ]
+        },
         "ProvisionActivity": {
           "@id": "ProvisionActivity-chips",
           "@type": "fresnel:Lens",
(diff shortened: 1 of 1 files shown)


source/vocab/display.jsonld
```diff
@@ -120,6 +126,12 @@
           "classLensDomain": "Identifier",
           "showProperties": [ "value", "typeNote", "marc:hiddenValue" ]
         },
+        "ISBN": {
+          "@id": "ISBN-chips",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "ISBN",
+          "showProperties": [ "value", "qualifier" ]
+        },
```
niklasl: Could be better to just add qualifier to Identifier chips.

source/vocab/display.jsonld
```diff
@@ -108,6 +108,12 @@
           "classLensDomain": "Title",
           "showProperties": [ "mainTitle", "title", "subtitle", "titleRemainder", "partNumber", "partName", "hasPart" ]
         },
+        "KeyTitle": {
+          "@id": "KeyTitle-chips",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "KeyTitle",
+          "showProperties": [ "mainTitle", "qualifier" ]
+        },
```
niklasl: Could be better to just add qualifier to Title chips.

source/vocab/display.jsonld
```diff
@@ -150,12 +162,24 @@
           "classLensDomain": "Country",
           "showProperties": [ "prefLabel" ]
         },
+        "marc:AddedEntryHierarchicalPlaceName": {
+          "@id": "AddedEntryHierarchicalPlaceName-chips",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "marc:AddedEntryHierarchicalPlaceName",
+          "showProperties": [ "countryOrLargerEntity", "city" ]
+        },
```
niklasl: I figure this was necessary for some operative requirement? (Since we avoid marc-terms, especially classes, here.)

source/vocab/display.jsonld
klngwll: Qualifier is not really used in the other Title types whereas KeyTitle only uses mainTitle and qualifier. So I think the specificity serves a dual purpose here in also excluding noise.

source/vocab/display.jsonld
klngwll: I went back and forth with this one for a while, since there are in fact several Identifiers with qualifier. For now I chose to try this one out, First of all I prefer if we don't make the chips longer than we have to, and this one is so explicit in praxis. After studying the almighty shapes I felt qualifier in the other Identifiers witness more about noise and data which might have ended up in the wrong place. I know this can also be of use but I think we need to revisit Identifiers more properly (in the future!) and see this as a test and see if the feedback is good or not.

source/vocab/display.jsonld
klngwll: Yes for older material where name of the places have shifted over time this is an important field. It is also a reminder for us that we want linked geodata ;)

source/vocab/display.jsonld
niklasl: Sounds good! I agree then that this captures a meaningful distinction (and singles out qualifier for ISBN as something we could address in isolation (since it is reasonably redundant if the ISBN is on the correctly described instance...)).

source/vocab/display.jsonld
niklasl: Makes sense. (And I figure qualifier here is also something indicative of a more relevant distinction we might address in the future.)

source/vocab/display.jsonld
niklasl: Fair game then!

PR #155, opened 2019-11-19
Update types with definitions and pending

klngwll:
* Add category pending to types which should not be addable in GUI.
* Add skos:definitions to BF2 Work and Instance types.
* Add exactMatch to LC Issuance types.
* General cleanup of structure, old comments and bf1 references which no longer resolve.

Sorry for the unreadable mess, should have split up fixes and cleanup in different commits ¯\_(ツ)_/¯

### DIFF
--- a/source/vocab/things.ttl
+++ b/source/vocab/things.ttl
@@ -232,7 +232,6 @@
     owl:equivalentClass bf2:Cartography .
 
 :Dataset a owl:Class;
-    :category :pending ;
     rdfs:label "Dataset"@sv;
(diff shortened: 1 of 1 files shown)


source/vocab/things.ttl
```diff
@@ -9,8 +9,8 @@
 @prefix foaf: <http://xmlns.com/foaf/0.1/> .
 @prefix bibo: <http://purl.org/ontology/bibo/> .
 
-@prefix bf1: <http://bibframe.org/vocab/> .
 @prefix bf2: <http://id.loc.gov/ontologies/bibframe/> .
+@prefix lcvi: <http://id.loc.gov/vocabulary/issuance> .
```
niklasl: Trailing slash?

PR #163, opened 2019-12-20
Remove sameAs from display

niklasl:
Future consideration: treating/displaying `sameAs` specially, since it is about as core as `@id`.


oBlissing:
> Future consideration: treating/displaying `sameAs` specially, since it is about as core as `@id`.

As in adding it to the `hiddenProperties` list in the GUI?

PR #164, opened 2020-01-10
Use prefLabel instead of label in Nationality

olovy:


niklasl:
But before we merge, and optionally add that, we must use infer: true for these everywhere in marcframe.json where they appear, so we get the correct MARC out on revert from the embellished items.

PR #172, opened 2020-01-29
Replace Libris language codes with ISO-639-3 equivalents

olovy:


olovy:
https://jira.kb.se/projects/LXL/issues/LXL-2916

olovy:
> so language preflabels will need to be obtained from some other source, e.g. http://www.lexvo.org/,

:+1:

niklasl:
We need to see if we use langCode (coerced to dc:ISO639-2) incorrectly. That is, our SPARQL construct for languages might make assumptions about which language list a code comes from which is now incorrect. (It is already incorrect for our "9"-codes, so we need to change the rule a bit. (E.g. keep "code" for *all*, but use langTag and langCode (and a new langCode3(?!) for the "code by list" properties...)

We should also add links to lexvo and wikidata. This PR is usable as is, as long as we follow-up with these formal requirements.

We could load and cache language lists (see e.g. https://github.com/libris/definitions/blob/develop/examples/queries/select-wikidata-languages.rq) and use that in the construct, to both extend our sources, make a union rather than a selection based on "our" list, and add the source link (kbv:derivedFrom for the data and skos:closeMatch for the entity) for all.

PR #179, opened 2020-02-18
Add properties used on some Items to display card

niklasl:
## Rationale
Sometimes Items are used to carry content-related information which for various reasons is not added to the proper Work or Instance.

## Checks
- [x] Built vocab with datasets.py


niklasl:
> LGTM!
> 
> Make Item-cards inherit from Item-chips while at it? Or do we want the properties in a different order?

I don't *think* the different order is intentional nor necessary. So I'll do that unless we can figure out why not.

PR #181, opened 2020-02-26
Add part as a kbv property

sarwik:
## Description
Add part as a kbv property able to add to Instance. This is a subsequent step of extracting part from isPartOf to the Instance ([LXL-1717](https://jira.kb.se/browse/LXL-1717)).

## Checklist:

### Tickets involved
[LXL-1717](https://jira.kb.se/browse/LXL-1717)

### Other PR involved
[librisxl 551](https://github.com/libris/librisxl/pull/551)

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -1091,8 +1091,9 @@
 
 :part a owl:DatatypeProperty;
     rdfs:domain :Instance ;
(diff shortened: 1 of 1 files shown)


sarwik:
> LGTM! Are the label and comment in line with praxis?

Updated label after discussions with Katarina and Elisabet. We test this label and change it if it is perceived as too long

PR #186, opened 2020-03-16
Make editionStatement/editionStatementRemainder repeatable

sarwik:
As a consequence of changing the mapping of bib 250 (not map to otherEdition when 250 is repeated) we now have to make editionStatement and editionStatementRemainder repeatable.

- [x] Built datasets.py


sarwik:
> LGTM, these are problaby in some templates and need updates as well.

Good catch! :) Handled with https://github.com/libris/lxlviewer/pull/605

PR #212, opened 2020-05-29
Add physicalDetailsNote

klngwll:
**Checklist:**

**Description:**
See https://github.com/libris/librisxl/pull/648 for details

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -923,7 +923,7 @@
 :physicalDetailsNote a owl:DatatypeProperty ;
     rdfs:label "Physical details note"@en, "Övriga fysiska detaljer"@sv ;
     rdfs:domain :Instance ;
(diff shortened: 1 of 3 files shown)


source/vocab/details.ttl
```diff
@@ -919,6 +920,13 @@
     rdfs:domain :Note;
     owl:equivalentProperty bf2:noteType .
 
+:physicalDetailsNote a owl:DatatypeProperty ;
+    rdfs:label "Physical details note"@en, "Övriga fysiska detaljer"@sv ;
+    rdfs:domain :Instance ;
+    rdfs:Range rdfs:Literal ;
```
sarwik, line 926: Do we usually add `rdfs:range` to `owl:DatatypeProperty`? But if we do, it should be with lowercase: `rdfs:range` 😄

PR #215, opened 2020-06-03
Update schemes.ttl

klngwll:

Addendum to https://github.com/libris/definitions/pull/213

### DIFF
--- a/source/schemes.ttl
+++ b/source/schemes.ttl
@@ -28,6 +28,8 @@
 
 # "ours"
 
+# TODO: properly link publishers in rdfs:comment
+
 <sao> a :TopicScheme;
     skos:notation "sao";
     dc:alternative "SAO";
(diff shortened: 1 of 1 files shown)


source/schemes.ttl
```diff
@@ -246,182 +250,299 @@
...
+<gtt> a skos:TopicScheme;
+    skos:notation "gtt";
+    dc:title "GOO-trefwoorden thesaurus"@en;
+    dc:title "GOO-trefwoorden thesaurus";
+    rdfs:comment "Nederländerna, Haag: Koninklijke Bibliotheek"@sv;
+    skos:exactMatch <http://id.loc.gov/vocabulary/subjectSchemes/gtt> .
+
+<humord> a skos:TopicScheme;
+    skos:notation "humord";
+    dc:title "HUMORD";
```
sarwik, line 316: Why not: `dc:title "HUMORD"@no, dc:title "HUMORD";` ?

klngwll: Was not sure because of the CAPITALS, I think we need to follow up on some of these soon anyway, so I think it is fine for now.

PR #221, opened 2020-08-20
Change domain of geographicCoverage to also include Instance

sarwik:
## Checklist:
- [x] I have built datasets

## Description
We will start mapping bib 052 to geographicCoverage (aligned with bf2) but also move it to instance (see https://jira.kb.se/browse/LXL-3293 for more info). This means that we need to change the domain for geographicCoverage to also include instance.

### Tickets involved
https://jira.kb.se/browse/LXL-3293

### PR involved
Changes of mapping and global changes of existing data: https://github.com/libris/librisxl/pull/694/files


klngwll (approved):
ITY! 😉

PR #225, opened 2020-09-07
Feature/add scheme description terms

klngwll:
## Checklist:

- [x] I have run datasets.py

## Description:

Add dc:alternative and foaf:page to context, for viewing schemes on id.kb.se


sarwik (approved):
LGTM!

PR #228, opened 2020-09-09
Add Class and category pending to swepub identifiers

klngwll:
## Checklist:
- [x] I have run datasets.py

## Description:
Not sure why we left these identifiers "hanging"? I guess if we want them to be properly usable on id.kb.se the owl:Class is needed (now they act quite broken). I added category:pending as advice to their status.

## Issues:
[LXL-3021](https://jira.kb.se/browse/LXL-3021)


sarwik (approved):
LGTM!

PR #236, opened 2020-09-14
Update about

klngwll:
## Description:
Update about document. Add text about relocation of model.

## Issues:
[LXL-3269](https://jira.kb.se/browse/LXL-3269)


sarwik (approved):
👌

PR #242, opened 2020-10-27
Add more TopicSchemes found in Libris data

olovy:


klngwll (approved):
LGTM! (with the knowledge that there are some general things we need to finetune in these descriptions later on.)

PR #243, opened 2020-10-28
Add TopicSheme ssif

olovy:


klngwll (approved):
LGTM. We should follow up on the code registering.

PR #244, opened 2020-10-29
Add specific ImageBitDepth chip because of value property

klngwll:
- [x] built vocab datasets.py

Add ImageBithDepth chip to display (usually DigitalCharecteristic things have labels, this one has value)


andersju (approved):
:ship:

olovy (approved):
:ship:

PR #245, opened 2020-11-09
Add lcschac scheme

klngwll:
- [x] I built definitions with datasets.py

Add lcschac scheme (Children's subject heading). Also needed for indicator 2 = 1 in JSONLD>MARC conversion.


olovy (approved):
LGTM!

PR #248, opened 2020-12-06
Add ISO639 datatypes and use in languages dataset

niklasl:
Fixes use of the datatypes as properties, and uses datatyped literals consequently in the table mapping and construct query.

Note that this changes the public language data a bit by renaming the specific properties involved in the fix.

Also note that the datatypes are consequently in the KBV namespace. This is mostly since two sub-datatypes of dc:ISO639-2 (for B and T) are not defined in DC Terms. (We may finally replace the mix of namespaces within the unified KBV namespace here, but that is a larger change.)

### DIFF
--- a/source/construct-languages.rq
+++ b/source/construct-languages.rq
@@ -11,15 +11,12 @@ construct {
         skos:prefLabel ?prefLabel ;
         skos:altLabel ?altLabel ;
         rdfs:comment ?comment ;
-        skos:notation ?not;
-        skos:notation ?langCode;
-        kbv:ISO639-2-B ?bibliographicLangCode;
-        kbv:ISO639-2-T ?termLangCode;
-        kbv:ISO639-3 ?iso_639_3;
+        skos:notation ?code;
+        skos:notation ?langTag ;
+        skos:notation ?langCode ;
         dc:isReplacedBy ?replacement;
         skos:exactMatch ?locLang ;
-        skos:closeMatch ?locVariant ;
-        skos:notation ?langTag .
+        skos:closeMatch ?locVariant .
 
 } where {
     graph <https://id.kb.se/dataset/languages> {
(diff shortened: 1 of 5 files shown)


source/construct-languages.rq
```diff
@@ -37,17 +34,22 @@ construct {
...
+            )
+            bind(iri(concat('http://id.loc.gov/vocabulary/iso639-2/', str(?locCode))) as ?locLang)
             graph <http://id.loc.gov/vocabulary/languages> {
-                optional {
+                {
                     filter(!contains(?prefLabel, "|"))
                     ?locLang skos:prefLabel ?prefLabel
-                }
-                optional {
+                } union {
```
niklasl, line 52: Using `union` appears more performant than `optional` when the contents is complementary in nature. (Which seems logical for join mechanics.)

olovy (approved):
:ship: 

Fixing Libris legacy MARC codes is still TODO but let's handle that separately.

i.e. for instance Meänkieli should be _one_ Language entity with ISO-639-3/langCodeFull 'fit' and URI https://id.kb.se/language/fit, but export as '9mk' in MARC.

Also, maybe we want sameAs URIs for 639-2-T (or T as main and B as sameAs)

klngwll (approved):
LGTM! There are some design choices we could discuss further, but a bit out of scope for now :)

niklasl:
So I assume we'll merge this as is to fix the erroneous use of a type as a property.

And separately continue with the issue of choosing which specific code list to use for the formal URI (albeit changing that should be done very carefully, as the current choice is based on some hard-fought assumptions of the MARC list being "right" on "our context", AFAICS).

More pressing might be a discussion and possible change to the way we use datatype-specific shorthands here, and naming thereof. Current choice of e.g. langCodeFull is based on the similar way we attempt to use "readable" properties (rather than e.g. ISO 8601 for dateTime information, and so on). That is, "intent" is preferred if possible. It is debatable here of course (as the "intent" is just "code", and the choice of list is in the datatype, whose naming is still "cryptic"). We need to determine which naming we can "live with", possible long-term if we don't revise this before the next release (as these codes are out there and eventually someone might depend on the shapes, semantics notwithstanding).

PR #249, opened 2020-12-10
Feature/fixing validation

klngwll:
First step in fixing some of our validation errors in marc properties. See each commit for details.

See [LXL-3421](https://jira.kb.se/browse/LXL-3421)

### DIFF
--- a/source/marc/enums.ttl
+++ b/source/marc/enums.ttl
@@ -1603,13 +1603,13 @@
 #     skos:prefLabel "Accompanying Matter (OBSOLETE)"@en,
 #         "Accompanying Matter (OBSOLETE)"@en .
 
-:VisualRunningTimeType a :CollectionClass ;
-    :category :pending ;
-    rdfs:subClassOf :EnumeratedTerm ;
-    skos:broadMatch kbv:Visual ;
-    skos:notation "VisualRunningTimeType" ;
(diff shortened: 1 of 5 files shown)


source/marc/enums.ttl
```diff
@@ -1603,13 +1603,13 @@
...
-    skos:notation "VisualRunningTimeType" ;
-    skos:prefLabel "Running Time"@en,
-        "Speltid"@sv .
+# :VisualRunningTimeType a :CollectionClass ;
+#     :category :pending ;
+#     rdfs:subClassOf :EnumeratedTerm ;
+#     skos:broadMatch kbv:Visual ;
+#     skos:notation "VisualRunningTimeType" ;
+#     skos:prefLabel "Running Time"@en,
+#         "Speltid"@sv .
```
olovy, line 1612: Sure we don't want to just erase it?

olovy (approved):
LGTM but I think it needs some more :eyes:

source/marc/enums.ttl
klngwll: Well we have taken the cautious path with these, since this one never had been used it might be less of a worry.

lrosenstrom:
I trust Fredrik more than myself on this. Looks reasonable!

lrosenstrom (approved):

PR #250, opened 2020-12-10
Fix broken links to loc for countries and relators

lrosenstrom:
Adds exactMatch with LOC links to country and relator definitions. For example Färöarna (https://libris.kb.se/katalogisering/0bvcdfl3251bd8st) gets
```
"exactMatch": [
   {
       "@id": "http://id.loc.gov/vocabulary/countries/fa"
   }
]
```
Also add english prefLabelByLang:
```
"prefLabelByLang": {
   "en": "Faroe Islands",
   "sv": "Färöarna"
}
```

### DIFF
--- a/source/construct-relators.rq
+++ b/source/construct-relators.rq
@@ -1,16 +1,16 @@
 prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
-prefix owl: <http://www.w3.org/2002/07/owl#>
 prefix skos: <http://www.w3.org/2004/02/skos/core#>
 prefix bf2: <http://id.loc.gov/ontologies/bibframe/>
+prefix madsrdf: <http://www.loc.gov/mads/rdf/v1#>
 prefix : <https://id.kb.se/vocab/>
 
 construct {
     ?idrel a bf2:Role;
         skos:notation ?notation;
         ?p ?o;
-        owl:equivalentProperty ?locrel;
+        skos:exactMatch ?locrel;
         rdfs:domain ?domain;
-        rdfs:subPropertyOf ?superrel .
+
 } where {
     graph <https://id.kb.se/dataset/relators> {
         {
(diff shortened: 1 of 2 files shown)


olovy (approved):
Works! 🔗

source/construct-relators.rq
```diff
@@ -1,16 +1,16 @@
...
 prefix bf2: <http://id.loc.gov/ontologies/bibframe/>
+prefix madsrdf: <http://www.loc.gov/mads/rdf/v1#>
 prefix : <https://id.kb.se/vocab/>
 
 construct {
     ?idrel a bf2:Role;
         skos:notation ?notation;
         ?p ?o;
-        owl:equivalentProperty ?locrel;
+        skos:exactMatch ?locrel;
```
niklasl, line 11: We should use `owl:equivalentProperty` (or add both if we need `skos:exactMatch` in our logic) as per the (rocket) discussion.

source/construct-relators.rq
```diff
@@ -1,16 +1,16 @@
...
         rdfs:domain ?domain;
-        rdfs:subPropertyOf ?superrel .
```
niklasl, line 13: Why drop this?

lrosenstrom: As with 'skos:notation',  'rdfs:subPropertyOf' is not available at https://id.loc.gov/vocabulary/relators.nt, so this code was not working. Hence nothing is dropped from a user perspective. However, I guess it would be nice to be able to link to the relators' superclasses, as was intended. In that case we need to get another relators dump or somehow query the individual relators to fetch this information(?).

source/construct-relators.rq
lrosenstrom: Sure. So which is preferrable, owl:equivalentProperty or both? Is there any particular reason relators should use this when e.g. languages such as https://libris.kb.se/katalogisering/766vj3jc919cchbh uses only exactMatch?

niklasl: I'd say ` owl:equivalentProperty`, but both can be used (for e.g. our display of these, and potentially queries or data consumers in general). (I'd prefer to declare ` kbvl:equivalentProperty` as a subproperty of both of these, but let's do that another day).

The reason is that these relators *are* properties (could be used as direct properties on entities), whereas languages are not.

source/construct-relators.rq
niklasl: Yes, I see. I can have a go at getting more of the data from their zipped dump (in the new branch https://github.com/libris/definitions/compare/feature/augment-loc-linking which also includes a modest speedup to the selection which we could cherry-pick if we need it now).

PR #252, opened 2021-01-23
Add model mapping to PREMIS

niklasl:

See [LXL-3569](https://jira.kb.se/browse/LXL-3569)


source/vocab/files-packages-representations.ttl
```diff
@@ -0,0 +1,121 @@
...
+
+# TODO: Rename to :isRepresentedBy, as PREMIS calls it (the label of :isr)?
+#:representedBy owl:sameAs :hasRepresentation .
+:hasRepresentation a owl:ObjectProperty ;
+    rdfs:domain :Embodiment ;
+    sdo:rangeIncludes :MediaObject, :Representation ;
+    owl:equivalentProperty relsubtype:isr ;
+    rdfs:subPropertyOf :associatedMedia .
+
+:associatedMedia a owl:ObjectProperty ;
```
sarwik, line 46: Already declared in relations.ttl. I will update the existing declaration.

sarwik (commented):
Great work! Some questions though... And more may come.

source/vocab/files-packages-representations.ttl
```diff
@@ -0,0 +1,121 @@
...
+    owl:equivalentProperty sdo:associatedMedia ;
+    rdfs:subPropertyOf :hasFormat .
+
+:includes a owl:ObjectProperty ;
+    sdo:domainIncludes :FilePackage, :MediaObject, :Representation ;
+    sdo:rangeIncludes :MediaObject, :Representation ;
```
marlip, line 52: Taking[ Premis' definition](https://id.loc.gov/vocabulary/preservation/relationshipSubType/inc.html) of the 'includes' relationship, the described and the referenced object have to be of different categories - so that a Representation 'includes' a MediaObject, but has a 'hasPart' relationship to other Representations. Would we not want to keep that distinction?  Or am I just missing a different relationship where Representation is the referenced object.

source/vocab/files-packages-representations.ttl
```diff
@@ -0,0 +1,121 @@
...
+    rdfs:domain :File ;
+    rdfs:range xsd:string ;
+    owl:equivalentProperty premis3:originalName ;
+    rdfs:subPropertyOf :label .
+
+:fileSize a owl:DatatypeProperty ;
+    rdfs:domain :MediaObject ;
+    #rdfs:range xsd:integer ;
+    skos:scopeNote "The value is an integer representing the file size in bytes, or a qualified string constisting of an integer followed by a conventional unit symbol (KB, MB, GB, TB)."@en ;
+    owl:equivalentProperty premis3:size, sdo:fileSize .
```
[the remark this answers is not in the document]
niklasl: Indeed, it should be `contentSize`! We reasonably want that name too (as `size` is too wide), albeit "content" is rather vague, so let's consider options ("mediaSize" or whatnot).

source/vocab/files-packages-representations.ttl
niklasl: Ah, great; I think I checked that and dropped it (we can update the existing declaration in this branch though, in case we'll backpedal on something).

source/vocab/files-packages-representations.ttl
niklasl: That makes sense! I _think_ I just overeagerly added `Representation` here. There is no use case for that, correct? (It doesn't look that way in the illustration on page 9 of premis-3-0-final.pdf either.)

source/vocab/files-packages-representations.ttl
```diff
@@ -0,0 +1,121 @@
...
+    rdfs:subPropertyOf :isPartOf .
+
+:parentPackage a owl:ObjectProperty; # TODO: how is it used?
+    rdfs:domain :FilePackage ;
+    rdfs:range :FilePackage;
+    rdfs:subPropertyOf :includedIn ;
+    skos:broader :inDataset .
+
+# TODO: find a better name? :creatingApplication? :toolUsed?
+:tool a owl:ObjectProperty ;
```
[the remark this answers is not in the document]
niklasl: I added it since we all seemed to use it in some of our examples (somewhere around `mix:ScannerCapture`, see the mets2kbv.xslt), but it can certainly be dropped in case we'd never use it.

source/vocab/files-packages-representations.ttl
marlip: Yes, you are right, I don't think there is a use case for that either. Great, then we have the same understanding.

source/vocab/files-packages-representations.ttl
sarwik: I agree. We will update in this branch :)

source/vocab/files-packages-representations.ttl
```diff
@@ -119,3 +119,13 @@ prefix : <https://id.kb.se/vocab/>
...
+
+:startTime a owl:DatatypeProperty ;
+    sdo:domainIncludes :MediaObject ;  #for event and bitstream?
+    owl:equivalentProperty sdo:startTime ;
+    rdfs:label "start time"@en, "starttid"@sv.
+
+:endTime a owl:DatatypeProperty ;
+    sdo:domainIncludes :MediaObject ; #for event and bitstream?
+    owl:equivalentProperty sdo:endTime ;
+    rdfs:label "end time"@en, "sluttid"@sv.
```
niklasl, line 127: Given that these apply to `MediaObject` in general, we might want to move them to details.ttl (or conversely move `MediaObject` with details here if we have any).

source/vocab/relations.ttl
```diff
@@ -51,7 +52,7 @@
     rdfs:label "has part"@en, "har del"@sv;
     rdfs:subPropertyOf :relatedTo;
     owl:inverseOf :isPartOf;
-    owl:equivalentProperty bf2:hasPart, dc:hasPart .
+    owl:equivalentProperty bf2:hasPart, dc:hasPart, relsubtype:hsp .
```
niklasl, line 55: We reasonably want to equivalently connect `:isPartOf` to `relsubtype:isp`.

Note that in both cases the (admittedly purely textual) restriction of "when these are the same object category" on the relsubtype properties makes me wonder if the more open and flexible `:hasPart` and `:isPartOf` are better as super-properties of these (and of `relsubtype:inc` and `relsubtype:isi` respectively).

However, in order to enforce such distinctions using OWL, we'd have to make restrictions on these per class. Perhaps keeping this as is and adding something like `:Representation rdfs:subClassOf [ a owl:Restriction ; owl:onProperty :hasPart ; owl:allValuesFrom :Representation ] .` and similarly for `:File` is enough. (And then, presumably, a similar pattern with `owl:allValuesFrom  [ owl:disjointWith :Representation ]` for `:includes` and :`includedIn`.)

niklasl: While we're at it, let's add `sdo:hasPart` and `sdo:isPartOf` equivalents.

source/vocab/files-packages-representations.ttl
```diff
@@ -0,0 +1,129 @@
...
+prefix dc: <http://purl.org/dc/terms/>
+prefix premis3: <http://www.loc.gov/premis/rdf/v3/>
+#prefix edm: <http://www.europeana.eu/schemas/edm/>
+prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubType/>
+prefix : <https://id.kb.se/vocab/>
+
+##
+# PREMIS 3.0
+
+:Instance rdfs:subClassOf premis3:IntellectualEntity .
```
sarwik, line 16: I'm thinking if `Creation` instead should be `premis3:IntellectualEntity`?

niklasl: Makes sense (meaning `:Creation rdfs:subClassOf premis3:IntellectualEntity`?).

source/schemes.ttl
```diff
@@ -865,3 +865,8 @@
     dc:alternative "UDK "@sv;
     rdfs:label "Universell decimalklassifikation"@sv .
 
+<aat> a :GenreFormScheme; # should we use ConceptScheme instead? aat as Scheme?
```
niklasl, line 868: Yes, it seems AAT contains concepts other than `:GenreForm`, so `:ConceptScheme` is more appropriate.

source/vocab/files-packages-representations.ttl
sarwik: I now see that my question was very vague. But yes exactly, as subClassOf :)

source/vocab/things.ttl
```diff
@@ -126,6 +126,7 @@
 #TODO move to concepts.ttl?
 :genreForm a owl:ObjectProperty;
     rdfs:label "genre/form"@en, "genre/form"@sv;
+    rdfs:domain :Representation;
```
niklasl, line 129: :genreForm already has `rdfs:domain :Endeavour` (from our automatic BF2 mappings), so this is unnecessary. (In fact it's too restrictive, since using `rdfs:domain` means the owning thing *must* be of that type, and not all things with `:genreForm` are `:Representation`s.)

datasets.py
```diff
@@ -187,6 +187,13 @@ def enumterms():
     return "/term/enum/", "2018-05-29T14:36:01.337Z", graph
 
 
+@compiler.dataset
+def aatterms():
+    graph = Graph().parse(str(compiler.path('source/aat-terms.ttl')), format='turtle')
```
niklasl, line 192: This file is missing? Also, we should probably not call this "aat", since we (probably) don't want to import all terms from AAT.

sarwik: Uuups... forgot to add it

sarwik: True. We will try to brainstorm something as comprehensible as possible. If you have any suggestions, we will be happy to accept them :)

source/vocab/things.ttl
sarwik: True! My bad. It was not my intention to just set `:Representation` as a domain. The purpose was to "expand" the domain to include both `: Representation` and`: Endeavor` (from the bf2 declaration). But nonetheless, `:Representation` is unnecessary here.

datasets.py
sarwik: Worth mentioning is that since we do not yet know how many terms we will add, we have so far added them directly to a turtle file, instead of constructing them with sparql

source/aat-terms.ttl
```diff
@@ -0,0 +1,21 @@
...
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix : <https://id.kb.se/vocab/> .
+@prefix aat: <http://vocab.getty.edu/aat/> .
+@base <https://id.kb.se/term/aat/> .
+
+
+<Page> a :GenreForm ;
+       skos:prefLabel "Page"@en, "Sida"@sv ;
+       :inScheme </term/aat> ;
+       :broader aat:300194222 .
```
sarwik, line 13: Or `closeMatch`? And I can see that `:inScheme </term/aat>` will be a problem when we don´t have `exaktMatch`.

sarwik:
We have now made some "final" adjustments and are ready to merge it to test how it works in Libris viewer. We have added some TODO's for the parts that are left to analyze/model/think about.

klngwll (approved):
A bit of testing in the dev-environments to check behavior including follow up on TODO:s. LGTM and it was nice to track rationale behind certain decisions in the comments. ⭐

PR #256, opened 2021-02-09
Output datasets to multiple JSON-lines files

niklasl:
This just adds some additional output json.lines-files per dataset, which we can try out for loading only specific datasets using the new dataset loader. 

If that works we can, when we want to, build and reload just specific parts. (E.g. by calling `python datasets.py languages`. See `python datasets.py -h` for more info.)


klngwll (approved):
ITY! Let's try

olovy (approved):
:ship:

PR #257, opened 2021-02-10
Add modified timestamp to datasets

kwahlin:
This adds a modified timestamp to each record representing a generated dataset. The timestamp corresponds to the modification time of the most recently updated file used in the generation of the dataset.

I'm not sure if the solution is sustainable. I wanted to avoid hard coding (which could be an option) the names of all the resources that a "dataset function" uses. Instead we look for file/directory name strings stated as constants in the functions. This works for now but we make assumptions that might not hold if/when those functions changes.

Also, the definitions dataset itself does not get a timestamp (should it?). This can of course be implemented, but would probably require a bit more shuffling in the code. Chose to skip it for now anyway.

### DIFF
--- a/datasets.py
+++ b/datasets.py
@@ -130,7 +130,9 @@ def vocab():
     vocab_created_ms = w3c_dtz_to_ms("2014-01-01T00:00:00.000Z")
 
     vocab_ds_url = urljoin(compiler.dataset_id, 'vocab')
-    compiler._create_dataset_description(vocab_ds_url, vocab_created_ms)
+    vocab_func = compiler.datasets['vocab'][0]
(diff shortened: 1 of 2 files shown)


niklasl:
I think I like this! It might border on "too clever", but it's succinct and possibly not too obscure (depending on audience of course).

What I have had in mind though, while just as internal, is _maybe_ a bit less advanced: we could "just" monitor calls to `compiler.path` 
(and rework uses of `self.cachedir` similarly) instead, to track used files. It would require instrumenting the `glob` method as well (monkey-patching the Path instance would probably do). It's less guesswork, but I'm not sure about the difference in complexity without trying it out though.

PR #260, opened 2021-02-14
Use a subclass of Path to track used resources

niklasl:
This is just a patch for #259 (and thus a feature branch on a feature branch). It avoids monkey-patching `Path` and also handles all tracking within the subclass, since all spawned instances from that are tracked as well (as `Path` cooperates by instantiating using the actual subclass).

Some things to note:
1. `Path` uses `__slots__` so instances aren't dynamically writable. We could just use a subclass to monkey-patch that, but I _think_ this patch ended up being more centralized and controlled.
2.  By using `set.clear()` the `Path` subclass only needs to be bound to the set, which I think is cleaner.
3. `Path` does some wonky initialization stuff (using `object.__new__`, so we couldn't do a standard `__init__` extension. Using `_init` works but isn't perfect, since it's an implementation detail (which might change in the future).
4. On the same note, it is also clever in choosing implementation when `__new__` is called, so to get the actual type I [used stackoverflow](https://stackoverflow.com/a/34116756) and just checked the type of an instance of Path.

### DIFF
--- a/lxltools/datacompiler.py
+++ b/lxltools/datacompiler.py
@@ -20,9 +20,7 @@ import sys
 import json
 import csv
 import time
-from os import path
 from datetime import datetime
-from functools import partial
 
 from rdflib import ConjunctiveGraph, Graph, RDF, URIRef
 from rdflib_jsonld.serializer import from_rdf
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
Nice! Concise and makes sense.

PR #261, opened 2021-02-15
Add inverseOf reproduction/supplementTo to Instance card

olovy:
Display reproductions and supplements without requiring two-way links.


klngwll (approved):
LGTM

PR #262, opened 2021-02-18
Feature/load library data

jannistsiroyannis:
Let's get Niklas script merged, so we can start scheduling this stuff.


olovy (approved):

libraries.py
```diff
@@ -12,8 +11,8 @@
         context='source/vocab-overlay.jsonld',
         record_thing_link='mainEntity',
         system_base_iri="",
-        union='libraries.jsonld.lines')
-
+        union='libraries.jsonld.lines',
+        created=datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc).isoformat()[:-6] + 'Z') # Beware, Compiler() demands UTC but treats as local time. Potential source of time bugs!
```
niklasl, line 15: The intent (as done in datasets.py) is to provide a fixed date (providing the created time for the outer dataset). Is there any reason not to here?

(Also, I'm aware of the off-by-timezone error in lxltools (blaming it on python 2 and more importantly my lazy fumbling), and have an incoming fix.)

olovy: Jannis is not working this week but I think his idea was to let the created timestamp reflect when it was updated in lddb (it would only change when the contents of the doc changed). That's what modified is for so it makes sense to set it to a fixed timestamp instead. We could use `date_created` from bibdb? https://bibdb.libris.kb.se/api/lib?sigel=S

niklasl:
I've changed the created time, added use of provided created and modified, and also added cache control plus removed more Python 2 cruft to boot (see individual commits for details).

I think it is in a good state  (I've ran both datasets.py and bibdb_datasets.py and glanced at the results), but please have a look prior to merge.

olovy:
We might want to to something smart in the DatasetImporter since the inital import of this leads to *a lot* of reindexing.

klngwll (approved):
LGTM! Checked out and tested local build which seem fine. Of course found some errors in the content but that is another matter :)

PR #263, opened 2021-02-21
Fix use of naive datetimes being based on local time

niklasl:
The `datetime` instances were naive (see the docs) and I made two errors based on the wrong assumption that these were *not* adjusted to local time:

* `datetime.fromtimestamp` is adjusted to local time unless a timezone is given. (Should have used `utcfromtimestamp` back in the Py 2.7 days, but here instead provides `timezone.utc` to be clear.)
* `time.mktime` is also based on local time. This is replaced with code keeping and using the timezone (but replacing `Z` with `+00:00` to keep our use of the former).

Those two errors cancel each other out in the formatted representation, but not internally in the milliseconds, which is offset by local time. Thus the `lxlslug` is based on milliseconds adjusted to local time, and not based on the UTC ("Zulu") time. To fix this here, I adjusted the fixed `created` timestamps per dataset, as I think it is better to just antedate the creation times (which are guesstimated minting times anyway and AFAIK not relied upon by anything else (yet)) in order to preserve the Libris XL IDs (not that I expect any external party to rely on them yet, but I don't want to tamper with cool URI principles).

Also, the truncating of micro- to milliseconds would fail when milliseconds where zero, due to `timespec='auto'` being the default! This has been fixed by properly specifying `timespec='milliseconds'`.

With this fix, modified timestamps are also properly represented in Zulu time (provided that the file system provides milliseconds from the POSIX epoch (in `Path.stat().st_mtime`), which is another assumption, albeit one that "should" hold...).

### DIFF
--- a/datasets.py
+++ b/datasets.py
@@ -45,7 +45,7 @@ def _get_repo_version():
 
 compiler = Compiler(base_dir=SCRIPT_DIR,
                     dataset_id=BASE + 'dataset/definitions',
-                    created='2013-10-17T16:07:48.000Z',
+                    created='2013-10-17T14:07:48.000Z',
                     tool_id=BASE + 'generator/definitions',
                     context='sys/context/base.jsonld',
                     record_thing_link='mainEntity',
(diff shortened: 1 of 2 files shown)


olovy (approved):
LGTM!

PR #265, opened 2021-02-26
Feature/vocab housekeeping2021 02

klngwll:

Some general fixups in vocab! See each commit for details.


kwahlin (approved):
LGTM!

PR #268, opened 2021-03-15
Feature/accessibility terms

klngwll:
- [x] I have run datasets.py

* Experimental and **pending** shapes for accessibility terms in bibliographic records for MTM usage. Faking linkable objecttypes locally is better than xsd:string for now... Lots of notes on continuation in dialogue with previous mentioned agent.


olovy (approved):
:ship:

PR #270, opened 2021-04-09
Add qualifier to Library and Bibliography from bibdb:dept

olovy:


jannistsiroyannis (approved):
I don't "get it", but it LGTM :P

oBlissing:
LGTM :)

oBlissing (approved):
LGTM :)

niklasl:
Might it be necessary to include the qualifier as a part of the `?library` IRI if bound? (Something like `if(bound(?qualifier), concat(?sigel, '/', ?qualifier), ?sigel)`...)

olovy:
> Might it be necessary to include the qualifier as a part of the `?library` IRI if bound? (Something like `if(bound(?qualifier), concat(?sigel, '/', ?qualifier), ?sigel)`...)

I don't think it is necessary since sigel is unique.
This is mainly so that they can be disambiguated when displayed.
e.g.
libraries:
sigel: Za, name: Chalmers tekniska högskola, dept/qualifier: Arkitektur- och samhällsbyggnadsbiblioteket
sigel: Zl, name: Chalmers tekniska högskola, dept/qualifier: Kuggen
bibliographies:
sigel ISTC, name: Kungliga Biblioteket, dept/qualifier: Inkunabler i LIBRIS
sigel SKTR, name: Kungliga Biblioteket, dept/qualifier: Skillingtryck
(these should maybe be renamed in bibdb since they're not really consistent with other bibliographies)

One thing we need to solve is when a sigel changes type between library and bibliography in bibdb. Because it gets a different mainId, it gets a different generated recordId. The datasetimporter sees it as a new record and the import fails because of a collision on sameAs. See LXL-3546.

klngwll (approved):
My first suggestion was actually the opposite, just concatenate it with the name, but can´t argue with the flexibility in presentation. Also treating Library class as somewhat abstract things as opposed to (possibly intersections/unions of?) their Organizations, qualifier seems a fair choice of allround property.

PR #271, opened 2021-04-15
Let itemOf and heldBy change places in display.jsonld

andersju:
Necessary for sorting (https://github.com/libris/librisxl/pull/888) to work slightly better.


olovy (approved):
:ship:

PR #287, opened 2021-05-27
Feature/linkable notation lists

klngwll:
Add RDA lists for:

MusicNotation
TactileNotation
MovementNotation (draft, not part of datasets build yet)

I have added these as separate files, each with a corresponding dataset and namespace under /term/rda/{list}. Are we fine with this design going forward? (For example; Carrier, Media, Content is in one set, but they also have a more direct connection which could make more sense).


source/movementnotation.ttl
```diff
@@ -0,0 +1,53 @@
...
+
+# MOVEMENT NOTATION linked to rdaregistry.com (No id.loc.gov counterpart 2021-05-27)
+
+<StepanovDanceNotation> a :MovementNotation ;
+    :label "Stepanovnotation"@sv ;
+    skos:notation "1001" ;
+    skos:exactMatch rdafnv:1001 .
+
+<Labanotation> a :MovementNotation ;
+    :label "Labanotation"@sv ;
```
olovy, line 19:
```suggestion
<LabanNotation> a :MovementNotation ;
    :label "LabanNotation"@sv ;
```

olovy (approved):
:ship: 

I think one file per category is fine.

sarwik (approved):
I think it becomes clear with one category per file. Ship it! 🚢

PR #288, opened 2021-05-27
Add encodingFormats and some premis/filePackage terms

sarwik:
* Add encodingFormats used by data.kb.se
* Add `hasIntellectualEntity` and `hasFilePackage`; needed by search in data.kb.se

### DIFF
--- a/source/encodingFormat-terms.ttl
+++ b/source/encodingFormat-terms.ttl
@@ -1,83 +1,95 @@
+@prefix : <https://id.kb.se/vocab/> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
 @prefix sdo: <http://schema.org/> .
 @prefix skos: <http://www.w3.org/2004/02/skos/core#> .
-@prefix : <https://id.kb.se/vocab/> .
 @prefix iana: <https://www.iana.org/assignments/media-types/> .
-@base <https://id.kb.se/term/imt/> .
+@prefix lcmenc: <https://id.loc.gov/vocabulary/mencformat/> .
+@base <https://id.kb.se/encodingFormat/> .
 
+# TODO: Add TAR
 
 <application/json> a :EncodingFormat ;
+    rdfs:label "JSON" ;
     skos:notation "application/json" ;
-    skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/json .
 
 <application/ld+json> a :EncodingFormat ;
+    rdfs:label "JSON-LD" ;
     skos:notation "application/ld+json" ;
-    skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/ld\+json .
 
 <application/marc> a :EncodingFormat ;
+    rdfs:label "MARC" ;
     skos:notation "application/marc" ;
-    skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/marc .
 
 <application/marcxml+xml> a :EncodingFormat ;
+    rdfs:label "MARCXML" ;
     skos:notation "application/marcxml+xml" ;
-    skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/marcxml\+xml .
 
+<application/msword> a :EncodingFormat ;
+    rdfs:label "MSWord" ;
+    skos:notation "application/msword" ;
+    skos:exactMatch iana:application\/msword .
+
 <application/pdf> a :EncodingFormat ;
+    rdfs:label "PDF" ;
     skos:notation "application/pdf" ;
-    skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/pdf .
 
 <application/rdf+xml> a :EncodingFormat ;
+    rdfs:label "RDF/XML" ;
     skos:notation "application/rdf+xml" ;
-    skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/rdf\+xml .
 
-<audio/mpeg> a :EncodingFormat ;
-    skos:notation "audio/mpeg" ;
-    skos:inScheme </term/imt> ;
-    skos:exactMatch iana:audio\/mpeg .
+<application/zip> a :EncodingFormat ;
+    rdfs:label "ZIP" ;
+    skos:notation "application/zip" ;
+    skos:exactMatch iana:application\/zip .
+
+<audio/mp3> a :EncodingFormat ;
+    rdfs:label "MP3" ;
+    skos:notation "audio/mp3" ;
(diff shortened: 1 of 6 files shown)


source/imt-terms.ttl
```diff
@@ -52,11 +57,26 @@
...
+    skos:notation "text/tab-separated-values" ;
+    skos:inScheme </term/imt> ;
+    skos:exactMatch iana:text\/tab-separated-values .
+
 <text/xml> a :EncodingFormat ;
     skos:notation "text/xml" ;
     skos:inScheme </term/imt> ;
     skos:exactMatch iana:text\/xml .
 
+<video/JPEG> a :EncodingFormat ;
```
sarwik, line 75: This should probably be `image/jpeg` instead, but I can´t find it among IANA´s media types: https://www.iana.org/assignments/media-types/media-types.xhtml 🤷

marlip: Jag, hittar inte heller. Men i I tabellen av imagetyper länkar iana till [denna](https://datatracker.ietf.org/doc/html/rfc2046) under 'jpeg':, och där i nämns jpeg flera gånger och [här](https://datatracker.ietf.org/doc/html/rfc2046#section-5.2.3) så ger de även ett exempel med just `/image/jpeg` så det verkar som att man köra på det.

source/imt-terms.ttl
```diff
@@ -52,11 +57,26 @@
...
+    skos:notation "video/JPEG" ;
+    skos:inScheme </term/imt> ;
+    skos:exactMatch iana:video\/JPEG .
+
```
marlip, line 79: Är det någon tanke att lägga till även ```<application/vnd.openxmlformats-officedocument.spreadsheetml.sheet> a :EncodingFormat ;
    skos:notation "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" ;
    skos:inScheme </term/imt> ;
    skos:exactMatch iana:application\/application\/vnd.openxmlformats-officedocument.spreadsheetml.sheet .
```
och
``` <application/msword> a :EncodingFormat ;        
    skos:notation "application/msword" ;        
    skos:inScheme </term/imt> ;                 
    skos:exactMatch iana:application\/msword . ```
    
    För excel arken som följer med ljudfilerna (om den gör det) och doc filerna som följer med Post- och inrikes tidningar tror jag att det var?

Comment (sarwik) [reply source/imt-terms.ttl:79]: Bra fråga. Jag funderade på dom oxå men blev osäker på om excel ska med för ljudfilerna iom att den ju bara innehåller info om vilka titlar varje fil har och SMDB-urlar? Tänker att den infon kommer vi ju ha i Libris? Men inte helt säker.
Likadant med `.doc` för Post- och inrikestidningar. Osäker om de ska in direkt nu i början? Men å andra sidan så är det ju ändå inte fel att lägga till excel och doc.

File: source/imt-terms.ttl [hunk 3]
```diff
@@ -37,6 +37,11 @@
     skos:inScheme </term/imt> ;
     skos:exactMatch iana:application\/rdf\+xml .
 
+<audio/mpeg> a :EncodingFormat ;
```
klngwll, line 40: I'm thinking perhaps we should have audio/mp3 and use broadMatch to ianas mpeg? Unless we mean /imt/ should strictly mean an exactMatch. Or else we can never express exactly as to which filetype it is since MPEG could mean mp1, mp2 and mp3. Alternatively use another list all together.

Both mp3 and jpeg is in this list also: https://id.loc.gov/vocabulary/mencformat.html altough some of the members seem a bit weird I thin we could align with that one as well, for the important ones.

```
MP3 (or mp3) as a file format commonly designates files containing an elementary stream of MPEG-1 Audio or MPEG-2 Audio encoded data, without other complexities of the MP3 standard. 
[...]
The Moving Picture Experts Group (MPEG) designed MP3 as part of its MPEG-1, and later MPEG-2, standards. MPEG-1 Audio (MPEG-1 Part 3), which included MPEG-1 Audio Layer I, II and III, was approved as a committee draft for an ISO/IEC standard in 1991
```
https://en.wikipedia.org/wiki/MP3

klngwll (changes requested):
I think we should add labels also :)

source/vocab/files-packages-representations.ttl
```diff
@@ -59,6 +60,11 @@
     owl:inverseOf :representationOf ;
     rdfs:subPropertyOf :associatedMedia .
 
+:hasManifest a owl:ObjectProperty ;
```
klngwll, line 63: Maybe even use a `:hasIIIFManifest` to be really explicit?

marlip:
We will need to support .zip and .tar files. Iana has an official media type for ZIP, but not for TAR. `application/x-tar` is commonly used (by [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) for example, or at [data.kb.se](https://data.kb.se/datasets/2020/10/swedish_nlp/spacy/) ), but is unofficial.
How do we approach that?  
@klngwll @sarwik

source/encodingFormat-terms.ttl
```diff
@@ -1,63 +1,111 @@
...
-@prefix : <https://id.kb.se/vocab/> .
 @prefix iana: <https://www.iana.org/assignments/media-types/> .
-@base <https://id.kb.se/term/imt/> .
+@prefix lcmenc: <https://id.loc.gov/vocabulary/mencformat/> .
+@base <https://id.kb.se/term/encodingFormat/> .
 
 
 <application/json> a :EncodingFormat ;
-    skos:notation "application/json" ;
+    rdfs:label "application/json"@en, "applikation/json"@sv;
```
klngwll, line 12:
```suggestion
    rdfs:label "JSON"
```

source/encodingFormat-terms.ttl
```diff
@@ -1,63 +1,111 @@
...
+    skos:notation "JSON" ;
```
klngwll, line 13:
```suggestion
    skos:notation "application/json" ;

source/encodingFormat-terms.ttl
klngwll: Btw, I am not sure we have added lang tags for abbreviations previously... Maybe we should? :)

marlip: Ah, there I got them switched up, well spotted! It was not intended to add lang tags on label :) Will push the corrections

marlip: Now that labels are abbreviations, you are right, seems that the lang tags are not necessary anymore. I will remove them, maybe we should not start with that in this very moment.

sarwik:
I have built datasets.py and it looks good with `/encodingFormat` instead of `/term/encodingFormat`. Any other comments @klngwll @olovy ? Or are we ready to merge?

PR #289, opened 2021-05-31
Feature/datakbse 187 add manifest terms

sarwik:
Add mapping to IIIF presentation API 3.0 https://iiif.io/api/presentation/3.0/ and to Web Annotation Data Model https://www.w3.org/TR/annotation-model/. 
For now we only have included the terms that we are using in data.kb.se (projekt "Öppna"), but more will be added based on the needs that projects "Tidningar" and "Visa" will identify.

- [x] I have built datasets


source/vocab/iiif-presentation-annotations.ttl
```diff
@@ -0,0 +1,118 @@
...
+@prefix dc: <http://purl.org/dc/terms/> .
+@prefix bf2: <http://id.loc.gov/ontologies/bibframe/> .
+@prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubType/> .
+@prefix : <https://id.kb.se/vocab/> .
+
+##
+# IIIF PRESENTATION API 3.0
+
+:id owl:equivalentProperty iiif_pres3:id .
+:type owl:equivalentProperty iiif_pres3:type .
```
olovy, line 18: Arent't these terms too generic and overloaded to be defined at the KBV "top level"?

source/vocab/iiif-presentation-annotations.ttl
```diff
@@ -0,0 +1,118 @@
...
+    # Lämplig superklass?
+
+:Image owl:equivalentClass :StillImage .
+:Video owl:equivalentClass :MovingImage .
+
+# TODO :Range a owl:Class ; --> Behov kommer med Tidningar och Visa
+# TODO :structures a owl:ObjectProperty ; --> Behov kommer med Tidningar och Visa
+# TODO :annotations a owl:ObjectProperty ; --> Behov kommer med Tidningar och Visa
+
+:items a owl:ObjectProperty ; # Begreppskrock med itemOf, hasItem och Item?
```
olovy, line 58: Krockar med
https://github.com/libris/definitions/blob/447e65ef899fd7f391ca9ce7695c9ac677dd58a7/source/vocab/platform.ttl#L371

source/vocab/iiif-presentation-annotations.ttl
niklasl, line 18: These are _probably_ not defined in their (non-RDF-documented)  vocabulary. I gather you've taken them from [their specific JSON-LD-context](https://iiif.io/api/presentation/3/context.json)? They are only defined there as aliases of  `@id` and `@type`. We (as in XL) do not use such aliases in the default context that _we_ associate with KBV (and use throughout in XL). Of course anyone using KBV from _another_ system-specific context are free to use any particular forms of JSON they need (ideally with motivations), as long as we keep the contexts separate (though one can build on another).

In any case, they should not be added as terms to KBV (nor as aliases to the XL default context at id.kb.se).

PR #290, opened 2021-06-01
Update DataCatalog and Dataset chips and cards

klngwll:
* Only extension from own chips stripping redundant work properties.
* Add relations.


olovy (approved):
:ship:

olovy (approved):
:ship:

PR #291, opened 2021-06-01
Add a11y enums and update vocab terms

klngwll:
* Add a11y enums
* Update vocab terms


klngwll:
Replaced by #526 and #529

PR #292, opened 2021-06-02
Remove place from library construct

klngwll:
Remove place from library construct for now, and replace uri > url


olovy (approved):
:ship:

PR #293, opened 2021-06-04
Change notation terms construction

kwahlin:
This adds prefLabels from RDAregistry to our notation terms.

A more generalized solution would be possible, however this consistent with the other datasets, i.e. each is constructed separately.

### DIFF
--- a/datasets.py
+++ b/datasets.py
@@ -191,15 +191,35 @@ def enumterms():
 
     return "/term/enum/", "2018-05-29T12:36:01.337Z", graph
 
+
 @compiler.dataset
 def musnotationterms():
-    graph = Graph().parse(str(compiler.path('source/musicnotation.ttl')), format='turtle')
+    graph = compiler.construct(sources=[
+        {
+            "source": Graph().parse(str(compiler.path('source/musicnotation.ttl')), format='turtle'),
+            "dataset": BASE + "dataset/musnotationterms"
+        },
+        {
+            "source": "http://rdaregistry.info/termList/MusNotation.nt"
+        }
+    ],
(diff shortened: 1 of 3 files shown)


kwahlin:
Do we want to add skos:definition as well?

olovy:
> Do we want to add skos:definition as well?

I would say yes!
We have it for [other RDA terms today](https://libris-qa.kb.se/katalogisering/t2g0vxm4w51qq5k6)
(noting that Swedish definitions are missing in the source)

olovy:
I don't think we should filter prefLAbel on lang sv?
In order to be consistent with other RDA terms.
In case we should filter: `sv` or `en`

kwahlin:
> I don't think we should filter prefLAbel on lang sv?
> In order to be consistent with other RDA terms.
> In case we should filter: `sv` or `en`

I added that filter to avoid getting two swedish prefLabels, where one (our own) begins with capital letter and the other (from RDAregistry) doesn't. If we don't want the filter we either have to change our own prefLabels to lowercase or add some kind of normalization. Not sure what's best.

olovy:
Ah ok, I read it the wrong way (filter everything but 'sv')

olovy (approved):
LGTM!

klngwll:
> Ah ok, I read it the wrong way (filter everything but 'sv')
> 
> Don't we want to use the RDA registry prefLabels instead, if present.
> @klngwll ?

For now it is easiest to go with our own defined, we can revisit this at a later stage but since it is a "note field" it is supposed to be written with a capitalized first letter anyway.

klngwll (approved):
LGTM

PR #294, opened 2021-06-07
Add more restrictions for range Agent

klngwll:
Restrict bibdb-agents from range on:
* seeAlso and hasVariant
* applicableInstitution and grantingInstitution


olovy (approved):
:ship:

PR #295, opened 2021-06-07
Add lens group search-cards to display.jsonld

olovy:
Used if available for ES indexing.

So that we can have things that are indexed but not displayed in cards.

* Add hasNote and summary for MTM
* Move shelfMarkStatus (already hidden by lxlviewer)


klngwll (approved):
LGTM!

andersju (approved):
:ship:

PR #296, opened 2021-06-07
Add term_status unstable to BibliographicAgent

klngwll:
Pending breaks functionality. Add vs_status to mark BibliographicAgent as an unstable term.


olovy (approved):
:ship:

PR #297, opened 2021-06-10
Remove duplicate definition of marc:Review

olovy:
Remove duplicate definition of marc:Review
Other definition [is here](https://github.com/libris/definitions/blob/22dfbcd49fdad150005e82725a64bb87777adb8f/source/marc/construct-enums.rq#L125)

Fixes this problem "Genre/form på verket: {Typ av innehållsbeskrivning/sammanfattning utan föredragen benämning}" caused by the the wrong definition being picked up from vocab.

https://libris-qa.kb.se/katalogisering/search/libris?q=%2a&%40type=Instance&_limit=20&instanceOf.genreForm.%40id=https%3A%2F%2Fid.kb.se%2Fmarc%2FReview

### DIFF
--- a/source/marc/indicators.ttl
+++ b/source/marc/indicators.ttl
@@ -239,8 +239,6 @@ marc:SummaryType a marc:CollectionClass;
     rdfs:subClassOf marc:EnumeratedTerm .
 marc:Subject a marc:SummaryType ;
     rdfs:label "Ämne"@sv .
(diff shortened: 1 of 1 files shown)


klngwll (commented):
Good catch! However we need to figure out what to do with those 5k summary types 🤔 there is a review / Review in bf2, but strangely enough LC has not used it. Would need to investigate this a bit.

PR #299, opened 2021-06-22
Feature/hub and work relations

klngwll:
For Work/Hub testing:

* Add Hub type
* Add representativeExpression relations

### DIFF
--- a/source/vocab/relations.ttl
+++ b/source/vocab/relations.ttl
@@ -2,6 +2,7 @@
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
 @prefix ptg: <http://protege.stanford.edu/plugins/owl/protege#> .
+@prefix vs: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
 
 @prefix prov: <http://www.w3.org/ns/prov#> .
 @prefix dc: <http://purl.org/dc/terms/> .
@@ -15,6 +16,7 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
 @prefix bflc: <http://id.loc.gov/ontologies/bflc/> .
 
 @prefix rdau: <http://rdaregistry.info/Elements/u/> .
+@prefix rdaw: <http://rdaregistry.info/Elements/w/> .
 
 @prefix : <https://id.kb.se/vocab/> .
 @prefix marc: <https://id.kb.se/marc/> .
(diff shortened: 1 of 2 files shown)


niklasl:
I *know* you say testing, but I'm diving in anyway for clarity (my own, posterity, and whatnot).

Has LC changed their minds on this? Originally the "hub" notion was only intended for the work "network", pretty much identical to how certain (SKOS) Concepts are considered top-concepts (see [`skos:hasTopConcept`](https://www.w3.org/2009/08/skos-reference/skos.html#hasTopConcept)) without having a special type.

I can see its editorial usefulness as a marker/restriction-class though, to be used either by means of documentation or *possibly* as a base-class of `:Work` (or worst-case, if absolutely necessary to guide cataloguers, an advanced design using limiting restrictions).

Remember though that what constitutes this "uniform" or "abstract/base/prototypical/sketch/concept" level of a Work *is different between specific subclasses*, e.g. `:Music` and `:Text`, and quite possibly different between different genres or other delineations. Also, recall that we have gone back and forth on whether *just* `:Work` is representative enough for the identity being on this abstract level. Since *if not*, we'd need "hub" variants for each subclass as well, or allow the use of multiple types.

Anyway, if needed, I *feel* "hub" ("nav") Ain't Good Enough™ a name, although I know coming up with the Good Enough™ name seems perpetually intractable. 😉 Maybe `:WorkHub` or `:HubWork` is better enough though (my main objection is that "hub" is way too generic, since can easily imagine "hub" topics, "hub" genres and so on if this method proves effective).

Finally, note that this reasoning is based on the assumption that we mean "hub" in the "central network nodes" notion of LC, and *not* as WEMI-works, as the latter comes with the additional complexity of requiring an abstraction where you cannot link from an `:Instance` directly to the "top". As LC says, "the data doesn't look like that", and we quite reasonably don't want to "graft" a huge amount of "anemic expressions" in between.

klngwll:
> I _know_ you say testing, but I'm diving in anyway for clarity (my own, posterity, and whatnot).
> 
> Has LC changed their minds on this? Originally the "hub" notion was only intended for the work "network", pretty much identical to how certain (SKOS) Concepts are considered top-concepts (see [`skos:hasTopConcept`](https://www.w3.org/2009/08/skos-reference/skos.html#hasTopConcept)) without having a special type.
> 
> I can see its editorial usefulness as a marker/restriction-class though, to be used either by means of documentation or _possibly_ as a base-class of `:Work` (or worst-case, if absolutely necessary to guide cataloguers, an advanced design using limiting restrictions).
> 
> Remember though that what constitutes this "uniform" or "abstract/base/prototypical/sketch/concept" level of a Work _is different between specific subclasses_, e.g. `:Music` and `:Text`, and quite possibly different between different genres or other delineations. Also, recall that we have gone back and forth on whether _just_ `:Work` is representative enough for the identity being on this abstract level. Since _if not_, we'd need "hub" variants for each subclass as well, or allow the use of multiple types.
> 
> Anyway, if needed, I _feel_ "hub" ("nav") Ain't Good Enough™ a name, although I know coming up with the Good Enough™ name seems perpetually intractable. 😉 Maybe `:WorkHub` or `:HubWork` is better enough though (my main objection is that "hub" is way too generic, since can easily imagine "hub" topics, "hub" genres and so on if this method proves effective).
> 
> Finally, note that this reasoning is based on the assumption that we mean "hub" in the "central network nodes" notion of LC, and _not_ as WEMI-works, as the latter comes with the additional complexity of requiring an abstraction where you cannot link from an `:Instance` directly to the "top". As LC says, "the data doesn't look like that", and we quite reasonably don't want to "graft" a huge amount of "anemic expressions" in between.

This is actually some of the questions that we at the moment are not really sure of and we need to discuss more with LC what the aim is turning out to be. Searching for bibframe hubs at id.loc.gov the picture becomes more complicated since they also use Work subclasses which both confirms the notion that hubs can be used anywhere and dispels the idea that it is a top concept layer.

See for example:

https://id.loc.gov/resources/hubs/0673e6e640f1696b34c1be0167f37e65.html

```
0 | "http://id.loc.gov/ontologies/bibframe/Work"
1 | "http://id.loc.gov/ontologies/bflc/Hub"
2 | "http://id.loc.gov/ontologies/bibframe/Audio"
3 | "http://id.loc.gov/ontologies/bibframe/NotatedMusic"
```

Since they don't have a URI as identifier I would say it is very much WIP and they are seem to be rendered two years ago, so there might have been things happening since. But it definitly asks the questions about what kind of relations from/to a hub would be possible.

I would also still consider the "Workhub" a conceptual level and leave Hub as a base class to be used for other things as you say.

PR #301, opened 2021-06-28
Broaden domain of url to also include Dataset

sarwik:
We will need to add `url` to `Dataset` for NB, auth and bibstat to be able to set correct `dcat:accessURL` (even though it belongs to `distribution`).

- [x] I have built dataset


klngwll (approved):
🚢

PR #305, opened 2021-07-22
Resolve doc paths relative to source directory

niklasl:
This is more explicit and handles a behavioural change between pathlib in Python 3.7 and 3.9.


andersju (approved):
Tested with Python 3.6 and 3.9. LGTM!

PR #306, opened 2021-08-13
Remove unused marc classes including related properties

klngwll:

Terms Not found in Marcframe, data or part of the definitions base.

Also see removal of mappings in https://github.com/libris/librisxl/pull/968


olovy (approved):
:wastebasket:

PR #307, opened 2021-08-19
Feature/remove unused marc properties

klngwll:

* Remove unlabelled marc properties.
* Comment out marc properties where the range of enums type neither exist or is used with another kbv property.

See also https://github.com/libris/librisxl/pull/968 for removed mappings.


kwahlin (approved):
LGTM!

PR #308, opened 2021-08-19
Feature/lxl 3654 add accession number

kwahlin:


klngwll (approved):
LGTM! The discussion we had on the domain of Identifier could be added to this PR or a separate depending on the solution.

PR #309, opened 2021-08-24
Lxl 3641 material

kwahlin:
This is a bit different from the bf2 implementation, where ranges are unnecessarily wide for our need. 
Should we also set `:Instance` as the domain for `:material` to be consistent with `:baseMaterial` / `:appliedMaterial` as they are? Or should we rather remove the domain from the latter and be more consistent with the bf2 implementation? Do we have a need to describe materials of other resource types than Instance?

### DIFF
--- a/source/vocab/enums.ttl
+++ b/source/vocab/enums.ttl
@@ -199,7 +199,7 @@
 
 :baseMaterial a owl:ObjectProperty;
     rdfs:domain :Instance; # Ta bort?
-    rdfs:range :BaseMaterial; # Ändra till Resource?
+    rdfs:range :Material; # Ändra till Resource?
     rdfs:label "bärande material"@sv, "Base material"@en;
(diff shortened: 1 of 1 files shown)


kwahlin:
Also: `materialOf` was omitted on purpose as I don't see a need for it.

klngwll:
> This is a bit different from the bf2 implementation, where ranges are unnecessarily wide for our need.
> Should we also set `:Instance` as the domain for `:material` to be consistent with `:baseMaterial` / `:appliedMaterial` as they are? Or should we rather remove the domain from the latter and be more consistent with the bf2 implementation? Do we have a need to describe materials of other resource types than Instance?

Possibly on Items but I don't think it has been used like that previously. Unique Items are still mostly described on Instance level, but maybe change the domain when there is a need.

kwahlin:
Added `:Instance` as domain for `:material` for consistency. As you say, we can widen the domain in the future if needed.

PR #310, opened 2021-08-26
Widen domain for identifiedBy

kwahlin:
This is mainly to make identifiedBy selectable for ImmediateAcquisition entities in the cataloging interface. 

We should put individual restrictions on different identifiers so that they get selectable only when the entity is of "right type": https://jira.kb.se/browse/LXL-3662


klngwll (approved):
LGTM

PR #311, opened 2021-08-26
Feature/lxl-3664-collection-arrangement

kwahlin:

See rationale in: https://github.com/lcnetdev/bibframe-ontology/issues/77

See marcframe changes: https://github.com/libris/librisxl/pull/978

Changes in the data not needed since removed/redefined terms have no occurences in Libris.

Set pending to arrangement/arrangementOf and Arrangement class for now.

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -1038,7 +1038,8 @@
     rdfs:label "Organization of material"@en, "Uppdelningsprinciper"@sv;
     rdfs:domain :CollectionArrangement;
     rdfs:range rdfs:Literal;
-    owl:equivalentProperty bf2:collectionOrganization .
+    owl:equivalentProperty bf2:collectionOrganization;
+    owl:sameAs :organizationOfMaterial .
 ##
 
 # Notes


klngwll (commented):
Maybe we should use owl:deprecated true on organizationOfMaterial, but it has never been actually in use right? 

Also check `bf-map.ttl` when the term is slightly different than the bf one (I usually do a search in the source folder for the term to smoke out any last remnants, for example in display and vocab overlay).

kwahlin:
@klngwll Good point about owl:deprecated, didn't think about that. However, since in this case the term doesn't occur in our data we just bin it?
I've removed occurrences of organizationOfMaterial in the source dir.

niklasl:
In *theory*, since we publish a public vocabulary (albeit with no contract in terms of stability or officiality) someone *might* use the term and be troubled by its sudden disappearance. In practise, this is extremely unlikely today, and since KBV is an application vocabulary tied to its data, we might even formalize that fact (e.g. in prose) and explicitly make no guarantees about terms other than that they are kept if used in our (KB:s) data.

In this case, if we're sure it's not used by anyone, we can certainly bin it.

If we *were* unsure of its external use, but considered the term as noise, leaving it commented out with a note like `# Dropped term 2021-09-08. Feel free to delete this comment note and term if encountered in 2025.`. That way its disappearance might "smoke out" usage through (negative) feedback. ;)

In general though, we should consider everything at `id.kb.se` as potentially used externally (even with us not knowing so), as we're an official agency with goals about long term stability and reusability. What we are lacking now is a published, formal contract along with documented practises for long term maintenance.

(I should probably rework this comment into a section in the README of this repository...)

kwahlin:
Great @niklasl, thanks for clarifying. I guess we can't be 100% sure that a term was never used externally. And as for the internal usage, I only checked the _current_ XL data.

In this case organizationOfMaterial is replaced by (or renamed to) collectionOrganization rather than just being dropped, should that also be communicated somehow?

klngwll:
> Great @niklasl, thanks for clarifying. I guess we can't be 100% sure that a term was never used externally. And as for the internal usage, I only checked the _current_ XL data.
> 
> In this case organizationOfMaterial is replaced by (or renamed to) collectionOrganization rather than just being dropped, should that also be communicated somehow?

I think this is a valid question if we should design a structured form of "isReplacedBy" or adding a "changeNote" to document this.

niklasl:
For terms with the exact same intent, I'd say `:newTerm owl:sameAs :oldTerm`, unless the `:oldTerm` is e.g. wrong or offensive, in which case we might just want to delete it. A sameAs allows for "semantic compatibility" if used in the wild, and also marks the term as "used", so we don't accidentally reinvent it.

PR #313, opened 2021-08-31
Pick up sigel registrant information from bibdb, when creating XL rec…

jannistsiroyannis:
…ords.

### DIFF
--- a/source/construct-libraries.rq
+++ b/source/construct-libraries.rq
@@ -39,7 +39,7 @@ construct {
 
     optional {
         ?bdb_id bibdb:libris_reg ?reg . filter ( ?reg = true )
-        bind (iri('http://id.kb.se/bibdb/registrant') as ?registranturl)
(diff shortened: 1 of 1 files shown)


source/construct-libraries.rq
```diff
@@ -36,6 +37,11 @@ construct {
         sdo:name ?name .
         # :organisation, sdo:address
 
+    optional {
+        ?bdb_id bibdb:libris_reg ?reg . filter ( ?reg = true )
+        bind (iri('http://id.kb.se/bibdb/registrant') as ?registranturl)
```
kwahlin, line 42:
```suggestion
        bind (iri('http://id.kb.se/term/bibdb/Registrant') as ?registranturl)
```
As per https://github.com/libris/definitions/pull/320/files

kwahlin (changes requested):

kwahlin (approved):

PR #314, opened 2021-09-01
Remove termComponentList from ComplexSubject chip

oBlissing:
Should solve the problem with data being doubled in the chips of `ComplexSubject`.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -325,7 +325,7 @@
           "@id": "ComplexSubject-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "ComplexSubject",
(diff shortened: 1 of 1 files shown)


klngwll (changes requested):
I don´t think it will. We removed prefLabels from the unlinked entities and they will then be left in the dark. We would need to add functionality that is something like this to work for all occurences:
`fresnel:alternateProperties ( :prefLabel, :termComponentList)`

olovy:
Here is an outline for how to handle prefLabel+termComponentList and formatting of termComponentList
https://jira.kb.se/browse/LXL-3631

oBlissing:
Alright, didn't realize they didn't have `prefLabel`.

PR #315, opened 2021-09-08
Add section about cleaning up terms

niklasl:
Principles and formulations up for debate!

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,12 @@ language labels more or less manually synced with various external origins. See
 It also contains a hand-curated set of RDF types extracted from MARC fixed
 field definitions (006, 007 and 008 for bib, auth and hold).
 
-### Vocab & MARC
+### Vocabulary Source Files
 
-The vocabulary is split into the formally deciced "vocab" terms, and the legacy
-"marc" (often unstable) terms stemming from MARC21 constructs not yet
-interpreted according to the new modelling principles (based on RDF and linked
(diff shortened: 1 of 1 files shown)


klngwll (commented):
Great addition and clarification! Might also be a good place to mention the usage of `:category :pending` or any other possible status flags like `vs:term_status "unstable"`?

niklasl:
@klngwll That's a good idea! I began drafting a paragraph, but I noticed we've used `vs:term_status "unstable"` a handful of times (lately?) instead of `:category :pending`. Any specific reason for that? As I recall, we've been using the latter form before since the use of strings that `vs:term_status` prescribes is poor practise.

niklasl:
@klngwll I added a section in abf31b0 to cover what we do now. We can revise that if we eventually decide upon one form for more things. (In theory, some parts could be computed from others, using vocabulary mapping techniques, but for now some extra book-keeping is probably quite manageable.)

klngwll:
> @klngwll That's a good idea! I began drafting a paragraph, but I noticed we've used `vs:term_status "unstable"` a handful of times (lately?) instead of `:category :pending`. Any specific reason for that? As I recall, we've been using the latter form before since the use of strings that `vs:term_status` prescribes is poor practise.

Yes the reason was to have no limitations in lxlviewer but still mark it as something that is not stable in usage. we could add a `:category :unstable` for this to keep the design as I think vs:term_status is a good idea but poorly implemented.

PR #317, opened 2021-09-09
Feature/lxl 3644 remove unused marc enums

kwahlin:
Remove superfluous marc enums that neither appear in our data or in marcframe. 

See also marcframe changes in https://github.com/libris/librisxl/pull/986


klngwll (approved):
LGTM! ⭐

PR #318, opened 2021-09-10
Add chip for ConceptScheme, containing only "title"

oBlissing:
Before this change, `ConceptScheme` lacks a proper chip, resulting in the fallback to `Resource`-chip, which contains multiple properties that doesn't need to be displayed for ConceptSchemes, resulting in overly verbose chips containing *both* the label, and the code.

Basically, this changes chips looking like
`Barnämnesord • barn` => `Barnämnesord`
`Tesaurus för grafiskt material • gmgpc/swe` => `Tesaurus för grafiskt material`

Smaller representations of these entities will still use the Token-definition for `ConceptScheme`, resulting in only the use of the `code`-property.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -85,7 +85,7 @@
           "@id": "ConceptScheme-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "ConceptScheme",
-          "showProperties": [ "prefLabel" ]
+          "showProperties": [ "title" ]
         },
         "ContentType": {
           "@id": "ContentType-chips",
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -81,6 +81,12 @@
           "classLensDomain": "MediaType",
           "showProperties": [ "prefLabel", "label" ]
         },
+        "ConceptScheme": {
+          "@id": "ConceptScheme-chips",
+          "@type": "fresnel:Lens",
+          "classLensDomain": "ConceptScheme",
+          "showProperties": [ "prefLabel" ]
+        },
         "ContentType": {
           "@id": "ContentType-chips",
           "@type": "fresnel:Lens",


oBlissing:
Just realized `ConceptScheme` doesn't use prefLabel. Let me get back on this :)

oBlissing:
One thing I want to be sure of before we merge this:

Will the `code`-property still be indexed and used in the search result data? We need it to display the token of the `ConceptScheme` "one step out" so to speak. But if I remember correctly, search results are based on cards, not chips, right?

Example:
I'm getting the search result for a Concept. In this resultdata, I'll have a ConceptScheme. In the chip for the Concept, we are displaying the prefLabel of the Concept, and the `code` of the ConceptScheme (which is the token for ConceptSchemes). Will the `code`-property still be available in the search result, even after we remove it from the chip version?

olovy:
> One thing I want to be sure of before we merge this:
> 
> Will the `code`-property still be indexed and used in the search result data? We need it to display the token of the `ConceptScheme` "one step out" so to speak. But if I remember correctly, search results are based on cards, not chips, right?
> 
> Example:
> I'm getting the search result for a Concept. In this resultdata, I'll have a ConceptScheme. In the chip for the Concept, we are displaying the prefLabel of the Concept, and the `code` of the ConceptScheme (which is the token for ConceptSchemes). Will the `code`-property still be available in the search result, even after we remove it from the chip version?

Unfortunately no. Index is card (and all properties with links) + one level of chips

We have "search-cards" for properties that should be indexed but not displayed but nothing similar for chips
https://github.com/libris/definitions/blob/1779613ddab8d59d146e37765e9dcb57366a0c79/source/vocab/display.jsonld#L993

oBlissing:
> We have "search-cards" for properties that should be indexed but not displayed but nothing similar for chips

We should maybe consider adding something like that. Or if we can solve this in some other way.

PR #319, opened 2021-09-15
Add more mappings to common vocabularies

niklasl:
These mappings are used to infer common forms (or "infer",  currently using [Target Vocabulary Maps](https://github.com/niklasl/trld#about-target-vocabulary-maps)).

More mappings will be added to this branch. Some of these may be "creative" as in not strictly reasonable, but possibly "useful enough", or just hand-waving depending on whose usage is considered. Opinions are very welcome! 🙂

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -199,13 +199,17 @@
 :isbn a owl:DatatypeProperty ;
     :category :shorthand;
     rdfs:label "ISBN";
-    rdfs:subPropertyOf :identifier ; #TODO? relate to https://schema.org/isbn
+    rdfs:subPropertyOf :identifier ;
+    owl:equivalentProperty sdo:isbn , bibo:isbn ;
+    owl:propertyChainAxiom ( [ rdfs:subPropertyOf :identifiedBy ; rdfs:range :ISBN ] :value ) ;
     rdfs:range :ISBN .
 
 :issn a owl:DatatypeProperty ;
     :category :shorthand;
     rdfs:label "ISSN";
-    rdfs:subPropertyOf :identifier ; #TODO? relate to https://schema.org/issn
+    rdfs:subPropertyOf :identifier ;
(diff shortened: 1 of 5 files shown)


source/vocab/details.ttl
```diff
@@ -946,6 +950,17 @@
...
+:datePublished a owl:DatatypeProperty;
+    :category :shorthand;
+    owl:equivalentProperty sdo:datePublished;
+    rdfs:subPropertyOf dc:issued;
+    owl:propertyChainAxiom ( :publication :date ) .
+
+:publisher a owl:ObjectProperty;
+    :category :shorthand;
+    owl:equivalentProperty dc:publisher, sdo:publisher;
+    owl:propertyChainAxiom ( :publication :agent ) .
```
klngwll, line 962: Maybe add a note that we also atm have ~600k linked publisher roles in contribution?

source/vocab/details.ttl
```diff
@@ -1254,6 +1272,11 @@
     owl:propertyChainAxiom ( :hasDuration :value );
     owl:equivalentProperty bf2:duration .
 
+:extentLabel a owl:DatatypeProperty ;
```
klngwll, line 1275: Maybe fits better as extentNote?

niklasl: Much better! fixed in efa01444a0b8e5f73321dfe5656b5fe643e78a02.

source/vocab/details.ttl
niklasl: *Maybe*, but only if our stated intent is to accept *and formalize* that odd (IMHO) way of describing publication (as being a kind of contribution to the work?!), and that the only place we'll attempt to address and normalize that is to be here, in this simple shorthand property.

And if that is to be the solution, what would the inverse reasoning be from e.g. a piece of DC or Schema.org description for a `:publisher` relation back to BF/KBV? Only a `:publication / :agent` chain? Or *also* a `:contribution / :agent` (both with `:role :publisher`)?

PR #322, opened 2021-09-27
Feature/datakbse 282 update described by

marlip:
Adding 'text/plain' to mediatypes (missing an exact match).

Updating domain and range of 'describedBy' for the changes made with https://jira.kb.se/browse/DATAKBSE-282?jql=text%20~%20%22describedBy%22


source/vocab/platform.ttl
```diff
@@ -61,8 +61,8 @@
     rdfs:range :Record .
 
 :describedBy a owl:ObjectProperty;
-    rdfs:domain :Work;
-    rdfs:range :Record;
+    rdfs:domain :Work, :File;
+    rdfs:range :Record, :FilePackage;
```
niklasl, line 65: These need to be changed to `sdo:domainIncludes` and `sdo:rangeIncludes` respectively (multiple ranges do, since the semantics of RDFS is "intersection of", so with the RDFS terms, the subject of a `:describedBy` relation would be *both* a `:Work` and a `:File`, etc.)

Alternatively, domain and range could both be `:Document`. In principle that is much better. Of couse, that'd enable the relation to be added on more entities in the cataloguing interface, but looking at marcframe, it already seems that it ought to be...

More pressingly though; how do you reckon a `:File` is `:describedBy` a `:FilePackage`, as opposed to being a `:includedIn` it?

niklasl (changes requested):
I have some concerns about the `:describedBy` alterations (see code comments).

source/vocab/platform.ttl
marlip: The reasoning behind a `:File` being `:describedBy` a `:FilePackage` is based on the fact that the metadata about the `:File`, such as `size`,  `checksum`, `filename` is part of the `:FilePackage`, and the description of the file and its features can be found there. Does that seem reasonable? 

From what I understood of the uri-debate, the solution we got to was that the `@id` of the `:File` would link to the mediafile itself. Therefore we needed to add a relation to `:File` in order to link it to its linked data description. If `includedIn` can be seen as doing just that, it could definitely be a (probably simpler) solution for us.

niklasl: I see. I thought the `:FilePackage` contained files, one of which was a metadata or manifest file in some RDF-serialization, within which these statements were serialized. In either case, it is correct that the `:describedBy` is to enable discovery of such triples.

For instance, if it is to be provided in a `Link` header from the file upon a `HTTP GET`, it should link to something with only descriptive metadata, and not a full file package. That is, presuming that that is the *archive file* and not just a *metadata document*), to save someone trying to get metadata for an image from downloading an entire archive. It is between these to resources, the file itself and the metadata document, for which there is an RDF serialization available (presumably at least as JSON-LD), that the `:describedBy` could be established (as a nice way of assigning a canonical record for the file).

It is also correct that the `@id`, that is, semantically, the IRI that *denotes the file*, must by definition lead to the file (based on the definition that a `:File` is a the representation itself in bytes, encoded in a MIME-type).

sarwik: Great. So then we are ok with the changes of `describedBy`?

sarwik: Short example of `FilePackage`:
```
{
  "@context": "https://id.kb.se/context.jsonld",
  "@id": "http://data-local.kb.se/dark-606554103",
  "@type": "FilePackage",
  "created": "2021-05-31T18:32:16.158000+02:00",
  ...
  },
  "includes": [
    ...
    {
      "@id": "http://data-local.kb.se/dark-606554103/bib11653825_17160626_145134_26_0001.jp2",
      "@type": "File",
      "fileName": "bib11653825_17160626_145134_26_0001.jp2",
      "encodingFormat": {
        "@id": "https://id.kb.se/encodingFormat/image/jp2",
        "@type": "EncodingFormat",
        "code": "image/jp2"
      },
      "describedBy": {
        "@id": "http://data-local.kb.se/dark-606554103"
      },
      "contentSize": 994482,
      "checksum": {
        "@type": "MD5",
        "value": "94c6ab0a2cc49aeeb692e6003bb46920"
      },
      "height": 1665,
      "width": 1035
    },
    ...
  ]
}
```

source/vocab/platform.ttl
```diff
@@ -61,8 +61,8 @@
     rdfs:range :Record .
 
 :describedBy a owl:ObjectProperty;
-    rdfs:domain :Work;
-    rdfs:range :Record;
+    rdfs:domain :Document;
```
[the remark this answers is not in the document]
niklasl: Exactly, we don't want to limit what can be described to just `:Document`s, but only `:Document`s can describe things.

niklasl (approved):
LGTM!

klngwll (approved):
LGTM

PR #325, opened 2021-10-18
Restructure contexts and add system contexts in new dataset

niklasl:
Redefine the vocab-overlay context as the core KBV context, and use a shared context for that and the current base context used when mapping external datasets.

Uses the JSON-LD 1.1 `@include` keyword *internally* when building the contexts and writing the new dataset for contexts.


datasets.py
```diff
@@ -2,11 +2,13 @@
...
 from rdflib import Graph, ConjunctiveGraph, RDF, Namespace
 from lxltools.datacompiler import Compiler, w3c_dtz_to_ms, last_modified_ms
-from lxltools.contextmaker import DEFAULT_NS_PREF_ORDER, make_context, add_overlay
+from lxltools.contextmaker import make_context, add_overlay
 from urllib.parse import urljoin
 
 
-NS_PREF_ORDER = DEFAULT_NS_PREF_ORDER + ['ldp']
+NS_PREF_ORDER = (
+        'bf2 bf madsrdf skos dc dctype prov sdo bibo foaf void ldp '
```
klngwll, line 10: add bflc?

olovy (approved):
:ship:

datasets.py
niklasl: Good call! Added in 121eacbcbcfe1a512fe779d81373c790a0e194e1

PR #326, opened 2021-10-19
Change administeredBy S -> Ssao for sao, saogf, barn, barngf, gmgpc, kao

olovy:
New sigel Ssao "Redaktionen för Svenska ämnesord"


olovy:
Yes, it's on the way. Pausing this until it's ready.

PR #337, opened 2021-11-04
Remove chips and tokens from nations/lang related classes

oBlissing:
All should be fine to just inherit from Resource now that we have alternateProperties on that.

* Removed chips for:
    * Language
    * Nationality
    * Country
* Removed tokens for:
    * Language.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -242,6 +242,14 @@
           "classLensDomain": "marc:AddedEntryHierarchicalPlaceName",
           "showProperties": [ "marc:countryOrLargerEntity", "marc:city" ]
         },
+        "Language": {
+          "@id": "Language-chips",
+          "@type": "fresnel:Lens",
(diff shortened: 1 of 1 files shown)


klngwll (changes requested):
At least Language have a card that extends on the chip. Are these calculated per Item, would it affect performance direct match vs traverse for things that have a big frequency in the hitlist/facet for example?

oBlissing:
> At least Language have a card that extends on the chip.

This PR is mostly about cleaning up code that's not needed.
I'm pretty sure that if we extend a specific chip, it needs to be explicitly defined (as in not just inherited). 
**I'll add a commit to this PR to add back the Language-chip.**

> Are these calculated per Item, would it affect performance direct match vs traverse for things that have a big frequency in the hitlist/facet for example?

They *should* cache their result, so if they're used a lot it shouldn't matter that much.

klngwll (approved):
LGTM

PR #338, opened 2021-11-04
Add contribution as alternative to hasTitle on Work chips

oBlissing:
Works quite nicely, but I would like some opinions on this before going ahead.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -123,7 +123,14 @@
           "@id": "Work-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "Work",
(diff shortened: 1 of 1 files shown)


klngwll:
> Works quite nicely, but I would like some opinions on this before going ahead.

The title and the (primary) contributor should be able to be viewed at the same time. It is an issue for the next sprint :)

oBlissing:
> The title and the (primary) contributor should be able to be viewed at the same time. It is an issue for the next sprint :)

Ok. Yes, maybe this is the wrong approach then. The main point of this was to have *something* for the Works without title, but we don't have to add that as an alternateProperty, we can just att `contribution` the normal way.

niklasl:
Should we close this, or rework it somehow?

PR #339, opened 2021-11-11
Feature/update display for identifiers

klngwll:
- Cluster identifiedBy and indirectlyIdentifiedBy (LXL-3650).
- Add ISSN chips and cards.
- Make alternateproperties for Identifier chips and add missing properties to cards.


olovy (approved):
:ship:

PR #340, opened 2021-11-11
Feature/integral relations

niklasl:
Adds the `:integral` collection and categorizes integral relations. Also adds the new  `:locatedIn` resource (which is defined as `:integral`).

### DIFF
--- a/source/vocab/relations.ttl
+++ b/source/vocab/relations.ttl
@@ -393,7 +393,6 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     owl:inverseOf :replacedBy .
 
 :hasReproduction a owl:ObjectProperty;
-    :category :integral ; # TODO: is this really the case, or is :reproductionOf enough?
     rdfs:label "Reproduced as"@en, "Reproducerad som"@sv;
     rdfs:domain :Instance;
     rdfs:range :Instance;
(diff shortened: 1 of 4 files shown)


olovy (approved):
LGTM!

source/vocab/relations.ttl
```diff
@@ -392,6 +393,7 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     owl:inverseOf :replacedBy .
 
 :hasReproduction a owl:ObjectProperty;
+    :category :integral ; # TODO: is this really the case, or is :reproductionOf enough?
```
klngwll, line 396: According to the comment on integral (nicely documented by the way 😃 ) it, I would not say a reproduction is necessery to understand the original resource.

klngwll (approved):
LGTM!

source/vocab/relations.ttl
niklasl: Probably not (and thanks 😊 ). What I had in mind was the perceived need to show the media of a reproduction alongside a print; but that is not *integral* information (there are lots of interesting things we do show which are not). OTOH it *might* be the case that `:associatedMedia` is `:integral` though, but we'll see.

I'll just remove the `:integral` from `:hasReproduction` for now and we'll see if any such need arises.

niklasl: Removed in  3fd9af2.

source/vocab/relations.ttl
```diff
@@ -393,7 +393,6 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     owl:inverseOf :replacedBy .
 
 :hasReproduction a owl:ObjectProperty;
-    :category :integral ; # TODO: is this really the case, or is :reproductionOf enough?
```
olovy: :+1:

PR #343, opened 2021-11-18
Add wikidata URIs to landskoder.tsv

olovy:
Unless we want to get them live in datasets.py?

Get MARC country code to URI
```
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX wdt: <http://www.wikidata.org/prop/direct/>
SELECT (replace(?code, "countries/", "") as ?countryCode) ?idd {
  ?idd wdt:P4801 ?code ;
  FILTER(strstarts(?code, "countries/"))
}

> wiki-countries.tsv
```

Add to landskoder.tsv
```
awk -F$'\t' -v OFS='\t' '
NR==FNR {
  cc[$1] = $2
  next
}

{
  print $1,$2,$3,$4,cc[$1]
}
' wiki-countries.tsv landskoder.tsv > landskoder2.tsv
```

### DIFF
--- a/source/landskoder.tsv
+++ b/source/landskoder.tsv
@@ -1,382 +1,382 @@
-code	label_sv	comment_sv	map_to_code
-9a	Åland	Librisdefinierad kod	
-9p	Palestina	Librisdefinierad kod	
-aa	Albanien		
-abc	Alberta 	"Se ""xxc"" (Kanada))"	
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
:clap: Should be good enough with this one-time effort?

PR #344, opened 2021-11-19
Update en/sv labels in kbv

klngwll:
* remove bf2 labels
* add missing sv/en labels in kbv terms

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -768,7 +768,7 @@
     rdfs:subClassOf :StructuredValue .
 
 :sourceConsulted a owl:ObjectProperty;
-    rdfs:label "Konsulterad källa"@sv;
+    rdfs:label "source consulted"@en, "konsulterad källa"@sv;
     rdfs:comment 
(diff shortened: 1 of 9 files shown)


source/vocab/details.ttl
```diff
@@ -943,21 +943,21 @@
 
 :otherYear a owl:DatatypeProperty;
     sdo:domainIncludes :PrimaryProvisionActivity ;
-    rdfs:label "kompletterande datum"@sv;
+    rdfs:label "other year"@en, "kompletterande datum"@sv;
```
kwahlin, line 946: Should be "kompletterande årtal" in Swedish?

kwahlin (approved):
See comment, but looks good otherwise!

source/vocab/details.ttl
klngwll: Could be! Added a note about it for now. I try to keep off content-changes which might require updated helptexts :)

kwahlin: Fair enough!

PR #345, opened 2021-11-19
Add UsageAndAccessPolicy dataset

niklasl:


source/policies.ttl
```diff
@@ -0,0 +1,17 @@
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+prefix : <https://id.kb.se/vocab/>
+base <https://id.kb.se/>
+
+</policy/freely-available> a :AccessPolicy ;
+    skos:prefLabel "Fritt tillgänglig"@sv, "Freely available"@en .
+
+</policy/gratis> a :AccessPolicy ;
+    skos:prefLabel "gratis"@sv, "gratis"@en ;
```
niklasl, line 9:
```suggestion
    skos:prefLabel "kostnadsfri"@sv, "gratis"@en ;
```
Thoughts/suggestions?

source/policies.ttl
```diff
@@ -0,0 +1,17 @@
...
+    skos:altLabel "free"@en .
+
+</policy/restricted> a :UsageAndAccessPolicy ;
```
niklasl, line 12: Restricted on usage and/or access, or just on access?

source/policies.ttl
klngwll: It might be important to specify OA context somewhere? Since it means "Öppen tillgång" or "Ej öppen tillgång". I'm gonna look around to see if we can find a preferred definition for this.

source/policies.ttl
```diff
@@ -0,0 +1,17 @@
...
+    skos:prefLabel "begränsad"@sv, "restricted"@en .
+
+</policy/cc0> a :License ;
```
klngwll, line 15: Maybe better to separate policy and license in the uri? (would need a change in the dataset setup as well ofc).

```suggestion
</license/cc0> a :License ;
```

source/policies.ttl
klngwll: From https://en.unesco.org/open-access/sites/open-access/files/215863e.pdf

"The other variant, where material is free to read but does not explicitly permit further types of re-use, is called  ‘gratis’ Open Access." - So I think gratis and restricted both can be seen as the Access condition?

olovy: I think these definitions need to be recorded in the term definitions, i.e. `description` and or `comment`

klngwll: #9923ae92a9abd6a705c67a641cdf11544c58d69b

source/policies.ttl
klngwll: Suggestion from the Swepub team applied # 9923ae92a9abd6a705c67a641cdf11544c58d69b

source/policies.ttl
niklasl: Comment out for now with a suggested change to `</term/cc/0>`, i.e. a separate dataset, getting these from creativecommons (or even better, cache records for those when external linked data lands)?

klngwll (approved):
🚢

PR #347, opened 2021-11-23
Make exactMatch a owl:SymmetricProperty

olovy:
So that `followReverseBroader` in XL can follow it it both directions.
It is a symmetric property and is defined as such in skos.

### DIFF
--- a/source/vocab/concepts.ttl
+++ b/source/vocab/concepts.ttl
@@ -117,7 +117,7 @@
     owl:equivalentProperty skos:definition ;
     owl:equivalentProperty madsrdf:definitionNote .
 
-:exactMatch a owl:ObjectProperty ;
+:exactMatch a owl:SymmetricProperty ;
     :category :integral ;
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
LGTM!

source/vocab/concepts.ttl
```diff
@@ -117,7 +117,7 @@
     owl:equivalentProperty skos:definition ;
     owl:equivalentProperty madsrdf:definitionNote .
 
-:exactMatch a owl:ObjectProperty ;
+:exactMatch a owl:SymmetricProperty ;
```
niklasl, line 120:
```suggestion
:exactMatch a owl:ObjectProperty, owl:SymmetricProperty ;
```

I think we have code checking if properties are (`owl:`)`ObjectProperty` both in the whelk and lxlviewer...

niklasl (approved):
LGTM!

olovy:
Let's park this one until we have time to investigate if we really handle multiple types on a property correctly everywhere.

PR #350, opened 2021-11-24
Suggest :publisher for :MediaObject

niklasl:


klngwll (changes requested):
I suggest we add the definition to be clear that the meaning of this property is much broader than the swedish ’förlag’ but an entity making the resource available (http://purl.org/dc/elements/1.1/publisher). I am still a bit uncertain about the providing part here if making available through hosting or repositories should be considered different in this context.

klngwll (approved):
LGTM!

PR #352, opened 2021-12-07
Add RDA-LC Material mapping

klngwll:
Mapping between RDA and LC Material.

* With swedish labels for those missing in rdaregistry.
* Possibly add swedish definitions later (in the works).
* sameAs to marc terms with significant usage (not many) but probably better to add more when analyzing and cleaning up existing data and/or revise MARC-mapping.
* For now does not handle terms that is exist in MARC or is used locally in Sweden but not in RDA, possibly align this with other schemes outside RDA.
* Add Material to display


kwahlin:
Nice! Some terms don't have Swedish prefLabel so maybe add that manually in material-terms.ttl? (Same with definition if we think it's needed.)

klngwll:
> Nice! Some terms don't have Swedish prefLabel so maybe add that manually in material-terms.ttl? (Same with definition if we think it's needed.)

Done. Definitions are not all translated so we could revisit them later.

source/material-terms.ttl
```diff
@@ -0,0 +1,68 @@
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix : <https://id.kb.se/vocab/> .
+@base <https://id.kb.se/term/rda/material/> .
```
niklasl, line 5: Or just `</material>`? And coordinate all the things (getty, maps/gmgpc...). 😉

source/material-terms.ttl
```diff
@@ -0,0 +1,68 @@
...
+<Ivory> a :Material ; skos:notation "1018" ; skos:exactMatch rdamat:1018, marcmat:ivo .
+<Lacquer> a :Material ; skos:notation "1019" ; skos:exactMatch rdamat:1019, marcmat:lac ; skos:prefLabel "lack"@sv .
+<Leather> a :Material ; skos:notation "1020" ; skos:exactMatch rdamat:1020, marcmat:lea .
+<MagneticParticles> a :Material ; skos:notation "1021" ; skos:exactMatch rdamat:1021, marcmat:fer ; skos:prefLabel "magnetiska partiklar"@sv .
+<Metal> a :Material ; skos:notation "1022" ; skos:exactMatch rdamat:1022, marcmat:mtl .
+# http://id.loc.gov/vocabulary/mmaterial/mix no RDA match.
+<Nitrate> a :Material ; skos:notation "1023" ; skos:exactMatch rdamat:1023, marcmat:nit .
+# http://id.loc.gov/vocabulary/mmaterial/nsf no RDA match.
+<OilPaint> a :Material ; skos:notation "1024" ; skos:exactMatch rdamat:1024, marcmat:oip ; skos:prefLabel "oljefärg"@sv .
+<Paper> a :Material ; skos:notation "1025" ; skos:exactMatch rdamat:1025, marcmat:pap ; :sameAs <https://id.kb.se/marc/Paper> .
```
niklasl, line 44: 👍 We must remove `marc:Paper` (to avoid id conflict in LDDB), remove the subclass `BaseMaterial` (in at least marcframe and deprecate in KBV). And then, map all other "BaseMaterial" marc-enums like this, here in this file (using `sameAs`).

And, of course test this by running MarcFrameConverter + revert and, crucially run a cleanup of all records linking to the marc-enums to re-link them to these terms.

source/material-terms.ttl
klngwll: If so I think we should either (1) not bring along RDA-definitions in the construct or (2) model the form to clearly indicate the source of each definition.

niklasl: We use `skos:exactMatch` here, and pick the english labels and definitions by choice. And presumably we will extend our labels with swedish definitions as well; we could also provide alterered english definitions and only fallback to the ones from the RDA sources).

Of course, we could augment this further, (using `derivedFrom` on the resulting records), but we haven't done so for the other interlinked datasets generated by `definitions` (such as languages).

klngwll: I started to think of using `exactMatch` and `closeMatch` as an indicator for proximity but that felt a little bit like semantic abuse. `derivedFrom` could of course be a possibility but becomes blurred if we have several sources. How is that qualifier Ontology coming along? :)

source/material-terms.ttl
klngwll: Commented out as TODO for now so no accidental merge will break things.

niklasl: We can set this PR as draft if we think we can address this here?

source/material-terms.ttl
niklasl: We use `exactMatch` for such qualifications already in definitions, and use that to "imply" that this is (may be) derived from the descriptions of the matches (as they are evidently "coordinated to" by us). Good enough? (We can always extend with clear record ("data document") provenance to state record-to-record relations; and in the future even star-annotations per piece if we want (as in need) to go there.)

source/material-terms.ttl
klngwll: Thinking about it I think it could be a good idea to separate the conversion alignments and cleanups into separate PR:s.

kwahlin:
Added mappings to terms in the AAT vocabulary in 8d983675514ce80b5230a00a5c4a5a947d76085d . Found the terms via this [sparql query](http://vocab.getty.edu/queries?toc=&query=select+distinct+%3Fmaterial+%3Flabel+%3FscopeNote+%3FparentString+%7B%0D%0A++++%3Fmaterial+gvp%3AbroaderExtended+aat%3A300264091+%3B%0D%0A++++++++++++++gvp%3AparentString+%3FparentString+%3B%0D%0A++++++++++++++gvp%3AprefLabelGVP+%5B+gvp%3Aterm+%3Flabel+%5D+.%0D%0A++++optional+%7B+%3Fmaterial+skos%3AscopeNote+%5B+rdf%3Avalue+%3FscopeNote+%5D+filter%28lang%28%3FscopeNote%29+%3D+%27en%27%29+%7D%0D%0A%7D%0D%0Aorder+by%28%3Flabel%29&implicit=true&equivalent=false) listing all terms that fall under ["Materials Facet"](http://vocab.getty.edu/aat/300264091). There were only a few terms that I couldn't find any match for, namely:
`<MagneticParticles>`
`<DiazoEmulsion>`
`<SilverHalideEmulsion>`
`<VesicularEmulsion>`
`<Synthetic>`
The last one possibly matches with ["synthetic fiber"](http://vocab.getty.edu/aat/300014058), however I think this is narrower?

Also coined several new terms with mappings to AAT in 13344b6b5730f106cf18d5fba1861dd104acaf54:
`<Bister>`
`<ColoredPencil>`
`<Crayon>`
`<FeltPen>`
`<Mixture>`
`<Panel>`
`<Papyrus>`
`<Pen>`
`<Pencil>`
`<Sepia>`
`<Silk>`
`<Wash>`

Up for discussion:
- The mappings for `<ColoredPencil>`, `<Crayon>`, `<FeltPen>`, `<Pen>` all fall under ["Tools and Equipment"](http://vocab.getty.edu/aat/300022238), are we comfortable making them instances of `:Material`?
- Is `<Crayon> skos:exactMatch <http://vocab.getty.edu/aat/300022415>` too broad to represent "färgkrita"? There is also the narrower ["pastels"](http://vocab.getty.edu/aat/300122621).
- No individual terms have been coined for "papp" and "kartong". Do we really need that? Or could `<Cardboard> skos:prefLabel "papp och kartong"@sv` be sufficient? If not, potential mappings in AAT are ["paperboard"](http://vocab.getty.edu/aat/300411320) and ["pasteboard"](http://vocab.getty.edu/aat/300162474) (which would be "papp" and which would be "kartong", I don't know).

klngwll:
> Added mappings to terms in the AAT vocabulary in [8d98367](https://github.com/libris/definitions/commit/8d983675514ce80b5230a00a5c4a5a947d76085d) . Found the terms via this [sparql query](http://vocab.getty.edu/queries?toc=&query=select+distinct+%3Fmaterial+%3Flabel+%3FscopeNote+%3FparentString+%7B%0D%0A++++%3Fmaterial+gvp%3AbroaderExtended+aat%3A300264091+%3B%0D%0A++++++++++++++gvp%3AparentString+%3FparentString+%3B%0D%0A++++++++++++++gvp%3AprefLabelGVP+%5B+gvp%3Aterm+%3Flabel+%5D+.%0D%0A++++optional+%7B+%3Fmaterial+skos%3AscopeNote+%5B+rdf%3Avalue+%3FscopeNote+%5D+filter%28lang%28%3FscopeNote%29+%3D+%27en%27%29+%7D%0D%0A%7D%0D%0Aorder+by%28%3Flabel%29&implicit=true&equivalent=false) listing all terms that fall under ["Materials Facet"](http://vocab.getty.edu/aat/300264091). There were only a few terms that I couldn't find any match for, namely: `<MagneticParticles>` `<DiazoEmulsion>` `<SilverHalideEmulsion>` `<VesicularEmulsion>` `<Synthetic>` The last one possibly matches with ["synthetic fiber"](http://vocab.getty.edu/aat/300014058), however I think this is narrower?
> 
> Also coined several new terms with mappings to AAT in [13344b6](https://github.com/libris/definitions/commit/13344b6b5730f106cf18d5fba1861dd104acaf54): `<Bister>` `<ColoredPencil>` `<Crayon>` `<FeltPen>` `<Mixture>` `<Panel>` `<Papyrus>` `<Pen>` `<Pencil>` `<Sepia>` `<Silk>` `<Wash>`
> 
> Up for discussion:
> 
>     * The mappings for `<ColoredPencil>`, `<Crayon>`, `<FeltPen>`, `<Pen>` all fall under ["Tools and Equipment"](http://vocab.getty.edu/aat/300022238), are we comfortable making them instances of `:Material`?
> 
>     * Is `<Crayon> skos:exactMatch <http://vocab.getty.edu/aat/300022415>` too broad to represent "färgkrita"? There is also the narrower ["pastels"](http://vocab.getty.edu/aat/300122621).
> 
>     * No individual terms have been coined for "papp" and "kartong". Do we really need that? Or could `<Cardboard> skos:prefLabel "papp och kartong"@sv` be sufficient? If not, potential mappings in AAT are ["paperboard"](http://vocab.getty.edu/aat/300411320) and ["pasteboard"](http://vocab.getty.edu/aat/300162474) (which would be "papp" and which would be "kartong", I don't know).

Great work! I Added some Wikidata links where it seemed to fit. I left out some that actually had matches on AAT/RDA/MARC identifiers since hey seemed ambigious or possibly broader/narrower so maybe revisit those sooner or later.

I would prefer for the equipment/tools to be it's own category of things rather than mash it into Material, but at least now we have a list matching a vocabulary.

source/material-terms.ttl
niklasl: Yes, this is for defining our sought-after material types, and a *separate* one for "folding" these and various MARC-forms together (FWIW). We might want to "SKOS-map" the MARC-enums here and now if the matches are obvious enough (to capture what we learn), and keep TODO:s about how some of them actually are `owl:sameAs` (like Paper here, which should now be `<Paper> ... ; skos:exactMatch ..., marc:Paper . # TODO: owl:sameAs marc:Paper`).

source/vocab/construct-enum-restrictions.rq
```diff
@@ -19,12 +19,12 @@ construct {
...
-(:baseMaterial	:BaseMaterial	:Microform	marc:FilmBaseType)
-(:baseMaterial	:BaseMaterial	:MovingImage	marc:FilmBaseType)
-(:baseMaterial	:BaseMaterial	:Visual	marc:ProjGraphEmulsionType)
-(:baseMaterial	:BaseMaterial	:StillImage	marc:NonProjectedType)
+(:baseMaterial	:Material	:Globe	marc:GlobeMediumType)
+(:baseMaterial	:Material	:Map	marc:MapMediumType)
+(:baseMaterial	:Material	:Microform	marc:FilmBaseType)
+(:baseMaterial	:Material	:MovingImage	marc:FilmBaseType)
+(:baseMaterial	:Material	:Visual	marc:ProjGraphEmulsionType)
+(:baseMaterial	:Material	:StillImage	marc:NonProjectedType)
```
klngwll, line 27: The drawback of changing these now is that we need to update all of those mappings in Marcframe also 🤔

kwahlin:
More or less good to go? More terms can be added later on, let's only include the ones we are are about now. 

There is still the problem with duplicate prefLabels from RDA/AAT. Beats me how to make source/construct-materials.rq pick prefLabel from AAT _only if_ there isn't one in that language in RDA. Maybe @niklasl wants to give it a try? We could also add English prefLabels in source/materials.ttl for terms missing in RDA and be content with that.

niklasl:
@kwahlin I can take a look next week! (Not sure how they can end up as duplicates if they're identical (given the set nature of triples)? 🤔 Is there a language-tag difference maybe?)

kwahlin:
@niklasl Thanks! You are right in that we wouldn't end up with duplicates _if_ the labels were identical. However, the RDA labels and the AAT labels differ in some cases, e.g. "paper" vs "paper (fiber product)".

niklasl (approved):
LGTM! :shipit:

PR #353, opened 2021-12-09
Remove topConcept from KBV ontology

klngwll:
Remove `skos:hasTopConcept` from KBV ontology.

### DIFF
--- a/source/vocab/things.ttl
+++ b/source/vocab/things.ttl
@@ -28,14 +28,6 @@
 @prefix marc: <https://id.kb.se/marc/> .
 
 
-# TODO: List top concepts to provide interfaces with:
-# - a set of starting points for e.g. navigation, creation
-# - a "roof" when finding base classes
-#: skos:hasTopConcept :Work .
-#: skos:hasTopConcept :Instance .
-##: skos:hasTopConcept :Print .
-
-
 ##
 # Creative Works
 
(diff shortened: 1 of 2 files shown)


source/vocab/things.ttl
```diff
@@ -28,14 +28,6 @@
...
 
 
-# TODO: List top concepts to provide interfaces with:
-# - a set of starting points for e.g. navigation, creation
-# - a "roof" when finding base classes
-#: skos:hasTopConcept :Work .
-#: skos:hasTopConcept :Instance .
-##: skos:hasTopConcept :Print .
-
-
```
niklasl, line 38: Is there anything hardcoded in the cataloguing client which would benefit from anything like this though? 🤔

klngwll: @oBlissing ?

oBlissing: Nope, no references at all to `hasTopConcept` in the UI code.

oBlissing: Did I understand the question? We DO have some similar things hardcoded:

https://github.com/libris/lxlviewer/blob/develop/lxljs/vocab.js#L169-L191

klngwll: @oBlissing Yeah the idea was if there was any function in for example viewer that would need something like that, we are still a bit unsure how this would translate to different contexts especially for 'Concept'.

niklasl: If we discover more than one context, we could define different `skos:Collection`s instead, listing classes of interest (for the then presumably documented contexts).

PR #354, opened 2021-12-10
Remove CataloguinRulesType-z

klngwll:
Remove Other as choice. Value z is default in marcframe for auth.

[LXL-1817](https://jira.kb.se/browse/LXL-1817)


kwahlin:
:ship:

kwahlin (approved):

PR #355, opened 2021-12-10
Add some additional generators.

jannistsiroyannis:
The URIs of these generators will be used in Libris history descriptions.


niklasl (approved):
LGTM!

PR #356, opened 2021-12-15
Feature/identifiers exact match

klngwll:
* Add exactMatch to [LC Standard Identifier Schemes](https://id.loc.gov/vocabulary/identifiers.html) members and wikidata Items for Identifiers in KBV.
* Also added notation to Identifier codes (also available from id.loc.gov) defined in marcframe for possibly more vocabulary-driven matching down the road .
* Some small cleanups and notes on further establishing pending Identifiers.

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -649,34 +649,42 @@
 
 :LibrisNumber a owl:Class;
     :category :pending ;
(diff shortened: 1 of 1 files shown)


niklasl (approved):
LGTM!

PR #363, opened 2022-01-21
Feature/caching sparql

kwahlin:
The changes in `datacompiler.py` makes it possible to send a construct query to a sparql endpoint and cache the result as a Turtle document. I think we should have this functionality since ready-made, well defined RDF documents aren't always available for download. In some cases external definitions can instead be retrieved via a sparql endpoint, however we wouldn't want to do that live every time we build the definitions, hence the cache. 

The other changes are to show how this feature can be used; AAT prefLabels are retrieved via http://vocab.getty.edu/sparql, cached and then added to the material terms. The current solution is not optimal as `construct-material-terms.rq` now adds all prefLabels from _both_ RDA and AAT which creates duplicates for some languages. I think we rather just want to complement with prefLabels from AAT when there are none from RDA. This should achievable by rewriting the construct in some clever way (I've tried really hard already but gave up in the end :smile:)

Remote querying requires the latest rdflib version which in turn requires Python 3.7. Could this be a problem?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ data used by the National Library of Sweden.
 
 ## Dependencies
 
-Requires Python 3.6+. (Use PyPy for a general speed improvement.)
+Requires Python 3.7+. (Use PyPy for a general speed improvement.)
 
 Preferably set up a virtualenv:
 
--- a/datasets.py
+++ b/datasets.py
@@ -224,6 +224,10 @@ def materialterms():
         },
         {
             "source": "http://rdaregistry.info/termList/RDAMaterial.nt"
+        },
+        {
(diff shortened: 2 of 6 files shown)


niklasl (approved):
Good stuff! I see no problem using latest RDFLib and Python 3.7+.

olovy (approved):
LGTM. We just need to make sure that devops tasks explicitly installs 3.7+ everywhere.

Also: Update definitions/README.md

kwahlin:
> LGTM. We just need to make sure that devops tasks explicitly installs 3.7+ everywhere.
> 
> Also: Update definitions/README.md

Handled in https://github.com/libris/devops/pull/107. Merging this.

PR #367, opened 2022-02-17
Merge :hasSeries into :partOfSeries

niklasl:
Also define domain and range and adjust labels.

### DIFF
--- a/source/vocab/relations.ttl
+++ b/source/vocab/relations.ttl
@@ -69,10 +69,6 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     rdfs:comment "Relaterad resurs vilken den beskrivna resursen är en fysisk eller logisk del av."@sv;
     owl:equivalentProperty bf2:partOf, dc:isPartOf, sdo:isPartOf, relsubtype:isp . # TODO: super-properties of relsubtype:isp instead?
 
-:partOfSeries a owl:ObjectProperty;
-    rdfs:subPropertyOf :isPartOf;
-    rdfs:label "part of series"@en, "del av serie"@sv .
-
 :references a owl:ObjectProperty;
     rdfs:label "hänvisar till"@sv;
     owl:equivalentProperty dc:references .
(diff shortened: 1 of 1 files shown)


source/vocab/relations.ttl
```diff
@@ -446,7 +447,7 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     rdfs:domain :SeriesMembership ;
     rdfs:subPropertyOf :object ;
     sdo:rangeIncludes :Instance ;
-    rdfs:label "ingår i serie"@sv .
+    rdfs:label "medlemskap i serie"@sv .
```
klngwll, line 450: Why change label here? for me it introduces a confusion with seriesMembership label.

niklasl: It came from a note I had (from some chat IIRC) about differentiating `inSeries` and `partOfSeries` through its label. But if we don't need to I'll gladly revert this label, since when _used_ its context (i.e. the type of the thing it's placed on  (`:SeriesMembership`)) should make the simpler label obvious, and when listed the domain/range context is reasonably needed anyway.

source/vocab/relations.ttl
```diff
@@ -424,11 +420,16 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     rdfs:subPropertyOf :accompanies;
     owl:inverseOf :supplement .
 
-:hasSeries a owl:ObjectProperty;
-    rdfs:label "har serie"@sv;
+:partOfSeries a owl:ObjectProperty;
```
klngwll, line 423: `.partOfSeries` is missing the corresponding `owl:inverseOf :seriesOf;` which in turn probably is best to change from `:hasSeries` to `:partOfSeries`.

source/vocab/relations.ttl
klngwll: Sounds good. We should do a follow up on that note on any possible differences, and if so document that in a comment or scopenote at least.

PR #368, opened 2022-02-25
Add isIssueOf relation

klngwll:


source/vocab/relations.ttl
```diff
@@ -444,6 +444,13 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     owl:inverseOf :hasSeries;
     owl:equivalentProperty bf2:seriesOf .
 
+:isIssueOf a owl:ObjectProperty;
+    # :category :pending ?
+    rdfs:label "Is issue of"@en, "nummer av"@sv;
+    rdfs:subPropertyOf :isPartOf;
+    # owl:inverseOf :hasIssue; ?
```
niklasl, line 451: We could have use for the reverse for providing a label, but that reverse should be pending or "abstract" unless we have a need to allow both directions (which does complicate consuming this data, as it would be a "denormalized" situation).

source/vocab/relations.ttl
```diff
@@ -444,6 +444,13 @@ prefix relsubtype: <http://id.loc.gov/vocabulary/preservation/relationshipSubTyp
     owl:inverseOf :hasSeries;
     owl:equivalentProperty bf2:seriesOf .
 
+:isIssueOf a owl:ObjectProperty;
+    # :category :pending ?
```
niklasl, line 448: I'd rather we allow this for "everybody", if it is what we will recommend for the newspaper data.

olovy, line 448:
```suggestion
```

sarwik:
LGTM. Does it have any relation to any of the properties of bibframe?

niklasl:
@sarwik There is no _specific_ relation equivalent to this in BF. This `:isIssueOf`is a specialization of `:isPartOf`, which is equivalent to`:bf:partOf` ([among others](https://id.kb.se/vocab/isPartOf)).

This means that without heuristics (e.g. some kind of declaration in KBV, perhaps using SCHACL, which a new [TVM](https://github.com/niklasl/ldtvm) pattern might pick up), there is no means to turn generic BF into this form. Specifically, thas means that an `<issue> bf:partOf <serial>` triple would not imply this specific relation, regardless of specificity of the respective types of the subject and object (since there could, formally, exist another part-of derived relationship between them).

With that said, the hypothesis we have here is that this specificity is valuable, since it "guides" the creation and usage of the data by clarifying that this is the specific partitive "issue" relation, without inspecting e.g. `issuanceType` on subject and object to come to that conclusion. A counter-argument would be that this specificity is redundant if the `issuanceType` implies it (but as the previous paragraph puts it, it is not necessarily enough information anyway, and it is certainly less immediate).

ebengtsson:
LGTM

source/vocab/relations.ttl
niklasl: For the "a reverse label would be nice" case, [it has been suggested](https://lists.w3.org/Archives/Public/public-rdf-wg/2012Nov/0333.html) that using a blank reverse with just the label is good enough. I like that pattern. When an actual need to use the inverse as a term in data arise, we can just mint a URI for it. Until then, it is handily "out of the way" as a bnode with a label.

klngwll: Fixed with https://github.com/libris/definitions/pull/368/commits/1416eb396e9a0a69136ec5fa57ebf2daa1c55c3e

PR #372, opened 2022-03-17
Skip altLabels in other languages on relators

kwahlin:
A bit overkill to collect altLabels in various languages, prefLabels (plus hiddenLabels where needed) are sufficient.


klngwll (approved):
LGTM

PR #374, opened 2022-03-18
Feature/bf2.1 fixes

klngwll:
See each commit for details.

NOTE: When testing domainIncludes and rangeIncludes mapping don't forget to check if term is overriden in `definitions/source/vocab/bf-map.ttl` like `kbv:referencedBy`

### DIFF
--- a/source/vocab/enums.ttl
+++ b/source/vocab/enums.ttl
@@ -5,6 +5,7 @@
 @prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 @prefix sdo: <http://schema.org/> .
 @prefix dbpo: <http://dbpedia.org/ontology/> .
+@prefix ptg: <http://protege.stanford.edu/plugins/owl/protege#> .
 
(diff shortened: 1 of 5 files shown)


source/vocab/enums.ttl
```diff
@@ -93,6 +94,17 @@
     rdfs:label "Frekvens"@sv ;
     owl:equivalentClass bf2:Frequency .
 
+:pubFrequency a owl:ObjectProperty;
+    # No relation in bf to frequency?
+    :category :pending ;
+    rdfs:range :PubFrequency ;
+    owl:equivalentProperty bf2:pubFrequency .
+
+:PubFrequency a owl:Class;
```
niklasl, line 103: Do we have some (marc) notion for this in the ... 008(?) mappings?

klngwll: Rationale for this addition is discussed here https://github.com/lcnetdev/bibframe-ontology/issues/76

We have an issue to add it properly later, this is just to hide it from the cataloging tool until then.

PR #376, opened 2022-03-29
Feature/mini geo ontology

kwahlin:


lxltools/contextmaker.py
```diff
@@ -133,6 +133,8 @@ def get_preferred(term, pred, ns_pref_order=None):
     relateds = list(term.objects(pred))
     term_pfx = _pfx(term)
     for related in relateds:
+        if type(related.identifier) == BNode:
```
niklasl, line 136:
```suggestion
        if isinstance(related.identifier, BNode):
```

niklasl:
We might (finally!) make use of this (or parts of, pending review), together with #475.

PR #381, opened 2022-05-04
Add description etc to qlit scheme

klngwll:
- [x] Run `datasets.py schemes`

### DIFF
--- a/source/schemes.ttl
+++ b/source/schemes.ttl
@@ -236,7 +236,7 @@
       
       Tesaurusen är framtagen inom projektet Databasen QUEERLIT (2021-2023) och innehåller ca 950 termer. Efter projektets slut övertas ansvaret för uppdateringar av KvinnSam - nationellt bibliotek för genusforskning.
     """@sv;
-    dc:title "Queer Literature Indexing Thesaurus"@sv, "Queer Literature Indexing Thesaurus"@en ;
+    dc:title "Queer Literature Indexing Thesaurus";
     dc:alternative "QLIT" .
 
 <prvt> a :TopicScheme;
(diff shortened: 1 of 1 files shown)


source/schemes.ttl
```diff
@@ -231,7 +231,13 @@
 
 <qlit> a :TopicScheme;
     skos:notation "qlit" ;
-    dc:title "Queerlit"@sv .
+    dc:description """
+      Queer Literature Indexing Thesaurus (QLIT) är ett kontrollerat, hierarkiskt ämnesordssystem med termer för indexering av skönlitteratur med HBTQI-teman. De flesta termer är mappade till exakta eller nära motsvarigheter i Svenska ämnesord (SAO), Barnämnesord (Barn), Library of Congress subject headings (LCSH) och/eller Homosaurus: an international LGBTQ linked data vocabulary (Homoit).
+      
+      Tesaurusen är framtagen inom projektet Databasen QUEERLIT (2021-2023) och innehåller ca 950 termer. Efter projektets slut övertas ansvaret för uppdateringar av KvinnSam - nationellt bibliotek för genusforskning.
+    """@sv;
+    dc:title "Queer Literature Indexing Thesaurus"@sv, "Queer Literature Indexing Thesaurus"@en ;
```
kwahlin, line 239: Is the Swedish title correct?

klngwll: Yes, I am glad you noticed :) this touches on a question about implementation and expectancy for data-users, not the least in our own interfaces. It is not in Swedish of course but it is also the name in Swedish. We have this "problem" with a lot of the schemes which are in native languages.

niklasl: If that's their official name in Swedish we can only point out that it could reasonably be called something in Swedish ("Tesaurus för indexering av Queer-litteratur"?). If a value is in "no" language, it makes sense to reduce redundancy by not repeating it for every possible language, and simply have one string without a language tag.)

PR #385, opened 2022-07-01
Add repeatable for used hasHold9XX fields and subfields

klngwll:
All marc:marc:hasHold9XX fields and subfields are repeatable according to https://katalogverk.kb.se/katalogisering/Formathandboken/Bestand/9XX/index.html

I have added only those appearing in our validation report of `Unexpected array. Key is not declared as repeatable in context` for hold to see if any new ones appear. Hopefully down the road these will be redundant anyway.


niklasl:
Since these are really just generated placeholders that should never see the light of day in usage, I'd rather work around them in the validation, or even filter them out. Declaring them in the context exposes them even more (when we ought to work to remove them). If these are ever to be explicitly used in the client, we need to mint proper terms for them.

niklasl:
I guess silencing them is problematic if we turn on the (here a bit overly strict, standards-wise) validation though. And if we get around to applying the context for normalizing on save, that would remove the wrapping array if there is only one value.

Could we _attempt_ go get rid of at least _some_ of these in one way or another, and only declare the ones that must be kept? (Ideally renaming them if so, though that might not give us so much more value...)

klngwll:
> I guess silencing them is problematic if we turn on the (here a bit overly strict, standards-wise) validation though. And if we get around to applying the context for normalizing on save, that would remove the wrapping array if there is only one value.
> 
> Could we _attempt_ go get rid of at least _some_ of these in one way or another, and only declare the ones that must be kept? (Ideally renaming them if so, though that might not give us so much more value...)

I agree we could take another round to check usage and relevance today. Leave this PR as a reminder for now.

PR #389, opened 2022-08-11
Display: Prefer KeyTitle and Title in Work/Instance hasTitle

olovy:
* Pick `KeyTitle` over `Title` over any `hasTitle` in cards and chips for `Instance` and `Work`.
* Display PrimaryContribution in Work chips

Depends on support for smarter `alternateProperties` in [lxlviewer](https://github.com/libris/lxlviewer/pull/813) and [librisxl](https://github.com/libris/librisxl/pull/1145).

Note that the title rule doesn't actually work if `range` is implemented correctly since it would match all subclasses, for example `:VariantTitle rdfs:subClassOf :Title`. 
Update: We decided to stick with the "broken" implementation since it matches out use cases.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -120,7 +120,14 @@
           "@type": "fresnel:Lens",
           "classLensDomain": "Instance",
           "showProperties": [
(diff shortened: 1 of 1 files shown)


niklasl (approved):
LGTM :shipit:

PR #391, opened 2022-09-07
Languages

olovy:
* Add Language to domain/range of broader/narrower]
* Update Language cards
* Add/update some languages

See individual commits

### DIFF
--- a/source/languages.ttl
+++ b/source/languages.ttl
@@ -2233,7 +2233,7 @@
     skos:broader <smi> .
 
 <smi> a schema:Language ;
-    rdfs:comment "Används restriktivt; se \"sia\" (akkalasamiska) \"sjd\" (kildinsamiska), \"sje\" (pitesamiska), \"sjk\" (kemisamiska), \"sjt\" (tersamiska), \"sju\" (umesamiska), \"sma\" (sydsamiska), \"sme\" (nordsamiska), \"smj\" (lulesamiska), \"smn\" (enaresamiska) och \"sms\" (skoltsamiska)"@sv ;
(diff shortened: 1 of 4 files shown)


niklasl (approved):
LGTM! If we have additional languages to link using `:broader` to aid in work matching, add these too. (Ping @kwahlin - I recall having seen some long-hanging fruits in a clustering example? Also @klngwll did you have any reference to LC mapping languages like this?)

source/languages.ttl
```diff
@@ -27,7 +27,8 @@
     rdfs:comment "-1500. \"9fs\" Librisdefinierad MARC-kod."@sv ;
     skos:notation "9fs",
         "9fs"^^kbv:LibrisLocalLanguageCode ;
-    skos:prefLabel "Fornsvenska"@sv .
+    skos:prefLabel "Fornsvenska"@sv ;
+    skos:hiddenLabel "svenska (fornsvenska)" .
```
niklasl, line 31: Add language tags on (all) the hidden ones too?

olovy: I think we could do it for the ones where it is obvious/unambigous.
Need to add `hiddenLabelByLang` to context though.

klngwll:
> LGTM! If we have additional languages to link using `:broader` to aid in work matching, add these too. (Ping @kwahlin - I recall having seen some long-hanging fruits in a clustering example? Also @klngwll did you have any reference to LC mapping languages like this?)

Looking into it now only was available as narrower terms on language groups like https://id.loc.gov/vocabulary/iso639-5/ine.html and exact matches to their subject headings from the 639-2 langs. So we can leave LC references for now.

source/languages.ttl
niklasl: Added that and some obvious tagged hidden values.

PR #393, opened 2022-09-13
Remove pending from URI identifier class

klngwll:
URI class is used and normalized to in all conversions for `024$2 uri` at this point.


kwahlin:
:ship:

niklasl:
Inclined to relabel it as "Uncool URI", but that wouldn't matter. :stuck_out_tongue_winking_eye: :shipit:

PR #395, opened 2022-10-28
Feature/include uri for resources

niklasl:
Let Resource-cards inherit from Resource-chips, which includes `uri` and fixes LXL-3961.

(This works locally. I _thought_ we used chips for these details in the index, but apparently not. And anyway it is generally good to inherit chips from cards unless special ordering is _required_. Both for making display consistent, and to ease maintenance.)


klngwll (approved):
LGTM

olovy:
> Let Resource-cards inherit from Resource-chips, which includes `uri` and fixes LXL-3961.
> 
> I _thought_ we used chips for these details in the index, but apparently not. And anyway it is generally good to inherit chips from cards unless special ordering is _required_. Both for making display consistent, and to ease maintenance.)

It's cards all the way for blank nodes. Chips for linked stuff.

PR #396, opened 2022-11-02
Adjust order in Person chip

olovy:
lifespan should come before title so that agents with the same
name and lifespan but different titles are next to each other when
sorting alphabetically:
Andersson, Lars, 1942-, pedagog
Andersson, Lars, 1942-, yrkesmilitär
    
(see LXL-3097)


niklasl (approved):
LGTM!

klngwll (approved):
LGTM

PR #397, opened 2022-11-02
Feature/new datasets

niklasl:
At this point, this is mainly for adding language-related resources used by the romanization feature and language/I18N logic in general. It also contains the first step of a dataset description improvement for usage in the new whelk dataset importer.

We could merge this now and follow up by a subsequent PR to fully land the dataset mechanics (rather than letting this hang indefinitely).


source/i18n/rules.ttl
```diff
@@ -0,0 +1,27 @@
...
+  :code "lessing" ;
+  :inCollection </i18n/rule/x0> .
+
+</i18n/rule/x0/skr-1980> a :LanguageTransformRules ;
+  :label "Translitterering enligt SKR 1980"@sv ;
+  :code "skr-1980" ;
+  :inCollection </i18n/rule/x0> .
+
+</i18n/rule/m0/iso-1995> a :LanguageTransformRules ;
+  :label "Translitterering enligt ISO 1995"@sv ;
```
olovy, line 15:
```suggestion
  :label "Translitterering enligt ISO 9:1995"@sv ;
```
I think we might need to model this in a different way.
The standard used here is [ISO 9](https://sv.wikipedia.org/wiki/ISO_9)
BPC47 extension T only has `iso` as a registered transform mechanism: [transform.xml](https://github.com/unicode-org/cldr/blob/main/common/bcp47/transform.xml)

[RFC6497](https://www.rfc-editor.org/rfc/rfc6497#section-2.5) mentions that other formats for `version` might be registered but they are not included in transform.xml.

So I think it is implicit from for example the tag `sr-Latn-t-sr-Cyrl-m0-iso-1995` that is ISO **9** because it is Cyrillic to Latin...

So maybe we can't map the m0 and x0 subfields directly to `LanguageTransformRules` if we want the rule definitions to be specific?

source/i18n/rules.ttl
```diff
@@ -0,0 +1,27 @@
...
+  :code "iso-1995" ;
+  :inCollection </i18n/rule/m0> .
+
+</i18n/rule/x0/btj> a :LanguageTransformRules ;
+  :label "Translitterering enligt BTJ"@sv ;
```
olovy, line 20:
```suggestion
  :label "Transkribering enligt Btj:s praxis för folkbiblioteken"@sv ;
```

source/i18n/rules.ttl
```diff
@@ -0,0 +1,27 @@
...
+  :code "btj" ;
+  :inCollection </i18n/rule/x0> .
+
+</i18n/rule/m0/alaloc> a :LanguageTransformRules ;
+  :label "Translitterering enligt ALA-LOC"@sv ;
```
olovy, line 25: "ALA-LOC" is both transliteration and transcription. "Romanisering enligt ALA-LOC" and more specific labels on `TransformedLanguageContent`?

olovy (approved):
Good progress!

See comments.

I think we should merge this now and then sort out:
- Modelling of transform rules
- Language tags for language with script, e.g. `mn-Mong` and `mn-Hang`. `LanguageForm`?
- Language tags and `sameAs` `/i18/lang` for `Language`s without ISO 639-1, e.g. `9mk`

source/i18n/rules.ttl
olovy: Or we just decide that `iso-1995` means "ISO 9:1995"

olovy: Some options are
- Denormalize everything and put labels and links to specs/romanization tables in TransformedLanguageForm
- Add an extra level of `LanguageTransformRule`s in some places. Relate with e.g. `broader`.
- Put a more specific label on some `TransformedLanguageForm` but describe all different rules in `LanguageTransformRule` for the broad ones like alaloc and iso

olovy: We want
- Specific labels like in https://metadatabyran.kb.se/beskrivning/specialanvisningar/mangsprak/romaniseringsscheman (maybe even more speific and better ones)
- m0 and x0 as entitites (?)

scripts/create_i18n_datasets.py
```diff
@@ -0,0 +1,136 @@
...
+
+ALA_LOC_NO_SCRIPT = ['am', 'chu', 'ka', 'hy']
+
+ALA_LOC_NON_SLAVIC_CYRILLIC = [
+    'abk', 'ady', 'alt', 'ava', 'bak', 'bua', 'che', 'chm', 'chv', 'dar',
+    'inh', 'kaa', 'kbd', 'kom', 'krc', 'krl', 'kum', 'lez', 'lit', 'nog',
+    'oss', 'rom', 'rum', 'rum', 'sah', 'sel', 'tut', 'udm', 'xal',
+]
+
+JUST_TAGS = ['ara', 'und']  # TODO: map lang to script
```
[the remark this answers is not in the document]
niklasl: Fixed in 989fab44f78496bdec4041f72f89d0b05568238a

source/i18n/rules.ttl
niklasl: Agreed. Label adjusted in 29e6837a78aa001fcee60b48f28a3559c983870c

source/i18n/rules.ttl
niklasl: Yes, we need to assess what this means. We *might* not need to remodel (or not much anyway). See the list at <https://unicode.org/reports/tr35/#37-unicode-bcp-47-t-extension>; reproduced here:

- `m0` - [Mechanisms](https://github.com/unicode-org/cldr/tree/main/common/bcp47/transform.xml) (typically authorities)
- `i0` - [Input Method Transformation](https://github.com/unicode-org/cldr/tree/main/common/bcp47/transform_ime.xml)
- `k0` - [Keyboard Transformation](https://github.com/unicode-org/cldr/tree/main/common/bcp47/transform_keyboard.xml)
- `t0` - [Machine Translation](https://github.com/unicode-org/cldr/tree/main/common/bcp47/transform_mt.xml)
- `x0` - [Private Use](https://github.com/unicode-org/cldr/tree/main/common/bcp47/transform_private_use.xml)

(There are also examples e.g. in <https://www.w3.org/International/multilingualweb/dublin/slides/23b-davis.pdf>, albeit that has obsolete links.)

Alas, I am a bit wary of directly using the Unicode.org XML data. I wrote an XSLT turning those into RDF we can use, but since [their license](https://www.unicode.org/license.txt) isn't CC0/PD but requires attribution, that reasonably applies to any results ending up at id.kb.se as well.

*If* we publish licenses for each Record, or the Dataset the record is in, clearly linking to documentation stating which data is derived from what, I believe we're good; based on the option:

> this copyright and permission notice appear in associated Documentation.

 But this is something we've yet to formally sort out. (Which we reasonably need for keeping "curated  cache records" of CC-licenses as well.)

source/i18n/rules.ttl
niklasl: Yes, it could we wise to extend the models for the `TransformedLanguageForm` (if you meant those?) to specify this.

olovy (approved):
LGTM

PR #398, opened 2022-11-08
Feature/extend romanization rules

niklasl:
This adds precision to the transformed language forms, to solve the conflation of `iso-1995` as the specific "ISO:9 from 1995, for Cyrillic to Latin script".

It is unclear if it is manageable and informative (or complicating) to do like this in general. At least in part, assessing this comes from how we use this data in the romanization feature.

If the rules are *just* informative support data, this is reasonably a good addition. But if we want to use these definitions in runtime algorithms (specifically to build BCP47 tags from `code` components), by this design, the algorithm would have to "climb up" via `broader` until if finds a code (though probably, as here, just one step). Might be OK, but it is more complex. For *maintaining predefined codes* I believe it is OK. But if we are to *generate* new `:TransformedLanguageForm`:s when the romanization API is called, we need to assess the complexity when using these definitions.

### DIFF
--- a/source/i18n/rules.ttl
+++ b/source/i18n/rules.ttl
@@ -1,6 +1,14 @@
 prefix : <https://id.kb.se/vocab/>
 base <https://id.kb.se/>
 
+</i18n/rule/m0> a :TermCollection ;
+  :code "m0";
+  :label "Standardiserade språktransformationsregler enligt BCP47"@sv .
+
+</i18n/rule/x0> a :TermCollection ;
+  :code "x0";
+  :label "Libris-specifika språktransformationsregler"@sv .
+
 </i18n/rule/x0/lessing> a :LanguageTransformRules ;
   :label "Translitterering enligt Ferdinand Lessings schema"@sv ;
   :code "lessing" ;
--- a/source/i18n/rules.ttl
+++ b/source/i18n/rules.ttl
@@ -19,10 +19,25 @@ base <https://id.kb.se/>
   :code "skr-1980" ;
   :inCollection </i18n/rule/x0> .
 
-</i18n/rule/m0/iso-1995> a :LanguageTransformRules ;
-  :label "Translitterering enligt ISO 9:1995"@sv ;
-  :code "iso-1995" ;
+</i18n/rule/m0/iso> a :LanguageTransformRules ;
+  :code "iso" ;
+  :label "Translitterering enligt ISO-standard"@sv ;
   :inCollection </i18n/rule/m0> .
+
+</i18n/rule/m0/iso-1995> a :LanguageTransformRules ;
+  :code "iso-1995" ;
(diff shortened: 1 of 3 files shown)


olovy (approved):
I don't have any better ideas.

niklasl:
I've added some more data to be used, and some comments on what we may still need (e.g. using the `libris-t-rules` collection as the source for what we can romanize).

source/i18n/rules.ttl
```diff
@@ -11,11 +21,47 @@ base <https://id.kb.se/>
   :code "skr-1980" ;
   :inCollection </i18n/rule/x0> .
 
+</i18n/rule/x0/kr76> a :LanguageTransformRules ; # :LanguageTranscriptionRules
+  :label "Translitterering enligt KR76 (Kjellberg)"@sv ;
```
olovy, line 25:
```suggestion
  :label "Transkribering enligt KR76 (Kjellberg)"@sv ;
```

source/i18n/rules.ttl
```diff
@@ -11,11 +21,47 @@ base <https://id.kb.se/>
...
+      :fromLangScript </i18n/script/Cyrl> ] ;
+  :seeAlso <https://metadatabyran.kb.se/download/18.6945cdaa174b74a2c361604/1601918194487/Transkriberingsschema_Kyrilliska%20alfabet_2015.pdf> .
+
+# TODO: form OK?
+</i18n/rule/x0/frhngbzs-1ed> a :LanguageTransformRules ;
+  :broader </i18n/rule/m0/alaloc> ;
+  :inCollection </i18n/rule/x0> ;
+  :languageFormSpecification [ a :TransformedLanguageForm ;
+      :inLangScript </i18n/script/Latn> ;
+      :fromLangScript </i18n/script/Cyrl> ] ;
```
olovy, line 39:
```suggestion
      :fromLangScript </i18n/script/Arab> ] ;
```

As far as I can tell there is no code for [persian alphabet](https://en.wikipedia.org/wiki/Persian_alphabet) in [ISO 15924](https://unicode.org/iso15924/iso15924-codes.html). `Arab` which in some sense "broader" seems to be used at least in some places around the web. Also an option to leave out `fromLangScript`?

niklasl: We should check with experts, but I *believe* the "broader sense" is at least informative enough and not formally wrong. In fact, according to Wikipedia, the [Persian alphabet](https://en.wikipedia.org/wiki/Persian_alphabet) is based on the [Arabic alphabet](https://en.wikipedia.org/wiki/Arabic_alphabet), where it is stated:

>  For the Arabic script as it is used by all languages, see [Arabic script](https://en.wikipedia.org/wiki/Arabic_script).

If I get that right, that means that Persian, written in the Persian alpabet, is in the Arabic script. Thus, with your fix, this is correct.

PR #400, opened 2022-11-09
Add chip for TransformedLanguageForm

olovy:
I guess we'd actually want prefLabel OR inLanguage + inLangScript here

something like 
```
          "showProperties": [ {"alternateProperties": ["prefLabel", ["inLanguage", "inLangScript"]] } ]
```
but that doesn't work

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -274,7 +274,9 @@
           "@id": "Notation-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "Notation",
-          "showProperties": [ "label", "prefLabel"]
+          "showProperties": [
+            {"alternateProperties": ["prefLabel", "label"]}
+          ]
         },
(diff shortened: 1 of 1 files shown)


niklasl (approved):
LGTM.

Agree that we want an `alternateProperties` though. I believe (given our context) that this is "legal fresnel":
```json
"showProperties": [
  {
    "alternateProperties": ["prefLabel", {"showProperties": ["inLanguage", "inLangScript"]}]
   }
]
```

At worst, the nested lens needs to be indirected via `fresnel:sublens`, like:
```turtle
<TransformedLanguageForm-chips> f:showProperties (
    [
      f:alternateProperties (
        :prefLabel
        [ f:sublens [ f:showProperties ( :inLanguage :inLangScript ) ] ]
      )
    ]
  ) .
```
(BTW, using Turtle to edit the display file might be nice...)

PR #401, opened 2022-11-09
Feature/type adjustments

niklasl:
See individual commits. These *should* be uncontroversial and agreed upon.

### DIFF
--- a/source/vocab/things.ttl
+++ b/source/vocab/things.ttl
@@ -57,8 +57,9 @@
     rdfs:comment "Abstrakt basklass för :Instance och :Item."@sv;
     rdfs:subClassOf :Endeavour .
 
-:Hub a owl:Class ; # LC Uses in conjunction with Work. SubClass a WorkHub and keep as baseClass in KBV?
-    # rdfs:label "Hub"@en, "Nav"@sv ;  - TODO: Naming things...
+:WorkHub a owl:Class ;
+    rdfs:label "Work Hub"@en, "Verkshubb"@sv ;  # TODO: Naming things...
+    skos:altLabel "Verksnav"@sv ;
     rdfs:subClassOf :Work ;
     owl:equivalentClass bf2:Hub .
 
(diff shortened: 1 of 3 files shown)


source/vocab/things.ttl
```diff
@@ -57,8 +57,9 @@
     rdfs:comment "Abstrakt basklass för :Instance och :Item."@sv;
     rdfs:subClassOf :Endeavour .
 
-:Hub a owl:Class ; # LC Uses in conjunction with Work. SubClass a WorkHub and keep as baseClass in KBV?
-    # rdfs:label "Hub"@en, "Nav"@sv ;  - TODO: Naming things...
+:WorkHub a owl:Class ;
+    rdfs:label "Work Hub"@en, "Verkshubb"@sv ;  # TODO: Naming things...
```
olovy, line 61:
```suggestion
    rdfs:label "Work Hub"@en, "Verksnav"@sv ;  # TODO: Naming things...
```

olovy (approved):
LGTM
See comment

klngwll (approved):
Is Classification decided? (referring to general discussions on making things less precise but still usable in different context with skos:Concept).

Not sure why we made DataDownload an `equivalentClass` in the first place...

source/vocab/things.ttl
```diff
@@ -57,8 +57,9 @@
...
+    skos:altLabel "Verksnav"@sv ;
```
niklasl, line 62:
```suggestion
    skos:altLabel "Verkshubb"@sv ;
```

niklasl:
`Classification` is decided insofar as these are not to be used as `subject` but as `classification` (in Swepub). So `Topic` is not desired as the type. De-typing all `Classification` terms as `Concept` is another challenge.

niklasl:
I'm going to keep "hubb" for now; we'll see in tests what crops up (perhaps hubs won't fly anyway).

PR #402, opened 2022-11-10
Add more hiddenLabels

kwahlin:
To be used for disambiguation when linking blank Language nodes in XL. See https://github.com/libris/librisxl/pull/1174


niklasl (approved):
LGTM (assuming that none of the hidden labels are used for heuristically identifying the language without further context, given that they're ambiguous (i.e. more than one language have the same hidden label))

olovy (approved):
:ship:

PR #403, opened 2022-11-21
Add langTag to language cards

lrosenstrom:
This is needed when adding language tags to language containers in the client.

However, 'code' is now displayed twice in cards in the search results.  
E.g. for Ukrainian:
```
Kod ukr
Kod uk
```


olovy (approved):
:ship:

PR #405, opened 2022-11-29
Add langTag to all non-local languages

kwahlin:
Had to split `construct-languages.rq` in two to get desired result from `coalesce(?langCodeShort, ?locCode)`. Don't think it's necessarily a bad thing though as future modifications will probably be easier. Now there is one query that operates on the iso639-1 graph and the other on the iso639-2 graph.

### DIFF
--- a/common.py
+++ b/common.py
@@ -167,7 +167,7 @@ def languages():
             "dataset": "http://id.loc.gov/vocabulary/iso639-2"
         }
     ],
-        query="source/construct-languages-1.rq")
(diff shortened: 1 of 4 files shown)


kwahlin:
@olovy [15acc5e](https://github.com/libris/definitions/pull/405/commits/15acc5ecc2a4e13f86a9cc8997072989a04b1589) should do the trick

olovy (approved):
:ship: :dash:

PR #406, opened 2022-12-16
Feature/newspaper terms

niklasl:
New terms used to describe newspaper serials (including new general temporal constructs).

### DIFF
--- a/source/vocab/details.ttl
+++ b/source/vocab/details.ttl
@@ -915,7 +915,7 @@
 
 :manufacture a owl:ObjectProperty;
     rdfs:subPropertyOf :provisionActivity;
-    rdfs:domain :Instance;
+    sdo:domainIncludes :Instance, :SerialEdition ;
     rdfs:range :Manufacture;
     rdfs:comment "Uppgift som identifierar tillverkningplats, tillverkarnamn, tillverkningstid."@sv;
     rdfs:label "manufacture"@en, "tillverkning"@sv .
@@ -1086,7 +1086,8 @@
     rdfs:label "upplageuppgift"@sv;
     rdfs:domain :Instance;
     rdfs:comment "En uppgift som identifierar den upplaga eller version till vilken en resurs hör, samt upplagespecifika upphovsuppgifter."@sv;
-    rdfs:equivalentProperty bf2:editionStatement .
+    rdfs:equivalentProperty bf2:editionStatement ;
+    owl:propertyChainAxiom (:isIssueOfEdition :label) .
 
 :editionStatementRemainder a owl:DatatypeProperty;
     rdfs:label "edition statement remainder"@en, "Påföljande upplageuppgift"@sv;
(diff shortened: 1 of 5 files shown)


klngwll (approved):
LGTM per our earlier discussions. I would not mind if we document a bit more notes on our intention. Otherwise let's put them to test asap!

source/vocab/details.ttl
```diff
@@ -1493,10 +1494,14 @@
...
+    rdfs:subClassOf :TemporalQualification ;
+    rdfs:label "temporal coverage"@en, "temporal täckning"@sv .
+
+:temporalCoverage a owl:ObjectProperty ;
     rdfs:label "Temporal coverage"@en, "Kronologisk täckning"@sv;
-    rdfs:domain :Work;
-    rdfs:range rdfs:Literal;
+    rdfs:domain :Work ;
+    rdfs:range :TemporalCoverage ;
     rdfs:comment "Kronologisk täckning avseende innehållet i resursen."@sv;
```
olovy, line 1505: "kronologisk" vs "temporal" täckning

source/vocab/enums.ttl
```diff
@@ -473,3 +475,58 @@
...
+    rdfs:subPropertyOf :pubFrequency .
+
+:FrequencyPeriod a owl:Class ;
+    rdfs:label "Frequency Period"@en, "Frekvensperiod"@sv;
+    rdfs:subClassOf :PubFrequency, :TemporalQualification .
+
+:firstIssueDate a owl:DatatypeProperty ;
+    rdfs:subPropertyOf :firstIssue, :startDate ;
+    rdfs:label "first issue date"@en, "första numrets datum"@sv;
+    sdo:domainIncludes :FormerTitle, :TemporalEntity ;
```
olovy, line 523: `:TemporalEntity` seems like something that is too abstract too have a `firstIssueDate`/`lastIssueDate`?
Should be `startDate`/`endDate`?

olovy (approved):
LGTM!

+1 on Fredrik's comment

source/vocab/enums.ttl
niklasl: Yes, this is an anomaly. It was either `SerialEntity` or generalize the date properties (and deviate from e.g. `PubFrequency`). But I'll make a consolidation attempt!

niklasl: Note though that is is not `TemporalEntity` that "has" these issue-dates; it is these issue-date-properties that are declared as describing a `TemporalEntity`. Which is still "over-reaching", but true. (It says nothing about "every temporal entity".)

source/vocab/details.ttl
niklasl: Addressed with "rationale" in c3e2e46645dcd75ebe7c4f044479bcb91719b8d4

niklasl:
I wrote some about the modeling/mappings in https://github.com/libris/signe2librisxl#about-the-mappings earlier. 

I could incorporate that into our definitions documentation going forward (my current plan being to reinvigorate the wiki here).

source/vocab/enums.ttl
niklasl: The point of these two is to specify (as a subproperty of start-/end-date), for the "temporal aspect" (here a newspaper serial), that the dates are given for issues (c.f. [`bf:firstIssue`](https://id.loc.gov/ontologies/bibframe.html#p_firstIssue) and `bf:lastIssue`). They can be put on multiple kinds of temporal entities, so i "land-grabbed" rather than cherry-picked them.

We can refine these later on, as `sdo:domainIncludes` is just a "usage hint". We should certainly come to a shared understanding and agreement upon the whole temporality notion of course. I added some more notes on that in ef2cff4b19e49943ece453a49c2fdb7348f8161e.

PR #407, opened 2022-12-19
More Script and TransformedLanguageForm definitions

olovy:
## Add more language `Script` definitions
For languages that there are ALA-LC romanization tables.
prefLabels taken from wikidata. Added Swedish where missing.

## Add more `TransformedLanguageForm` for ALA-LC
This should cover all ALA-LC romanization tables [1] that are
used in Libris [2] that we have `Language` definitions for.

[1] https://www.loc.gov/catdir/cpso/roman.html
[2] https://metadatabyran.kb.se/beskrivning/specialanvisningar/mangsprak/romaniseringsscheman


niklasl (approved):
LGTM! We want to regenerate `source/i18n/tlangs.ttl` since there is no build step defined for these.

(We could define such, but I'm thinking of reducing the need of those to a bare minimum, possibly in favour of running "maintenance scripts" (or strive to add more of these as regular Libris resources; _perhaps_ with a specific maintenance UI).)

niklasl:
Regenerated in b1cbf47fdfae1167b8ca3fe81f641b54ae2b7bf7

PR #409, opened 2022-12-21
Make it easier to manage Script labels

niklasl:
- Add SPARQL construct for getting them from wikidata
- Add SPARQL utility script for local use
- Update wikidata concept URI:s (their canonical forms use `http`)
- Add simple invocation example to top of file (for maintenance)


source/i18n/scripts.ttl
```diff
@@ -1,104 +1,108 @@
-# Start here: <https://en.wikipedia.org/wiki/ISO_15924>
+# For a list of available alphabets/scripts, start here:
+#   <https://en.wikipedia.org/wiki/ISO_15924>
+# When you add a new script or need to update all labels; get labels for all by calling:
+#   $ python3 scripts/construct.py source/remote/construct-matching-wd-labels.rq THIS_FILE
+# then add the resulting :prefLabel data here.
```
kwahlin, line 5: Maybe add a reminder that there should always be a `@sv`-tagged label

kwahlin (approved):
LGTM! (At least as a short-term solution)

olovy:
Nice!
I think we should pick the Swedish prefLabels from scripts.ttl to keep the under our control.

source/i18n/scripts.ttl
niklasl: Good point; added in 64c600f6316a95f0dbad9422dd8bb7eb64d160e6

PR #410, opened 2023-01-25
Add ShelfMarkSequence to domain and range of replacedBy

olovy:
To make it possible to add via the cataloging interface. 
Already set by the shelfmark year-bump script.


kwahlin (approved):
LGTM

klngwll (commented):
> Maybe add these properly as well with `:Creation`?
> 
> _Suggested use - With Work or Instance
> Suggested value - Work or Instance_
> 

Not used in Marcframe so leave as is for now.

klngwll (approved):
LGTM

PR #412, opened 2023-02-06
Add cataloguersNote to Item index

klngwll:
Where code for NBs veckolista can be found. For *testing*


niklasl (approved):
If necessary right now, let's do this! But let's also add an issue to FMT to adress this practise (at least clarify which various notes and markers are added to items, so they can be cohesively explained, or even better be consolidated, and possibly turned into links where appropriate).

PR #413, opened 2023-02-13
Tag platforms terms with `:category :platform`

niklasl:
and move library-things from platform to agents.

This is mainly house-keeping for clarity. We may eventually define these in a vocabulary of their own and alias them in the JSON-LD context. (But as we would like to make such a vocabulary more neutral (e.g. using `w3id.org`) and have some collaboration set up on that and the generic XL platform, let's defer those specifics until we know how that will pan out.)

### DIFF
--- a/source/vocab/agents.ttl
+++ b/source/vocab/agents.ttl
@@ -307,3 +307,33 @@
     rdfs:subClassOf [ a owl:Restriction;
             owl:onProperty :hasPart; owl:someValuesFrom :Jurisdiction ] .
 
+
+##
+# Libris administrative "things"
+
+# Possible idea forward with a base "operator" Class for bibdb things, for now we inherit Library and Bibliography from Agent and Collection.
+# :OperatorProfile a owl:Class ;
+#     :category :abstract ;
+#     rdfs:label "Sigel"@sv ;
+#     rdfs:subClassOf :Agent .
+
+:Library a owl:Class ;
+    rdfs:label "Library"@en, "Bibliotek"@sv ;
+    rdfs:subClassOf :Collection, :Agent .
+    # rdfs 
+
+:Bibliography a owl:Class;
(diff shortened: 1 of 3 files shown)


klngwll (commented):
Good to move out Library things to agent and a good start to figure out where these piecies fit in.

However I am not sure all administrative/meta things are (entirely) considered platform so maybe we need a better common understanding of what we mean by platform so we can define it properly.

My first thought were that terms which relates to bf2 should probably not be platform and I identified three piles of things to discuss.

Record/Adminmetadata - not quite platform in itself, but platform specific forms?
Dataset - certainly a grey zone, but overlap with bibliographic data.
Hydra/Query/Fresnel/API terms - definitely platform.

niklasl:
The "platform" category means "required by the platform". These terms may _also_ be "bibliographic", but the platform cannot work without them. If the "bibliographic" notions are 1:1 with the technical, the term is the same, otherwise we need subclasses (which may also derive from other subclasses). Until the system is ontology-aware (as opposed to the current, raw-JSON-lexlcal-term mechanism), a JSON-LD context then have to "bind" e.g. `Record` to  `some:SpecificBibliographicNotionOfRecord` (subclassing `PlatormRecord`).

klngwll:
> The "platform" category means "required by the platform". These terms may _also_ be "bibliographic", but the platform cannot work without them. If the "bibliographic" notions are 1:1 with the technical, the term is the same, otherwise we need subclasses (which may also derive from other subclasses). Until the system is ontology-aware (as opposed to the current, raw-JSON-lexlcal-term mechanism), a JSON-LD context then have to "bind" e.g. `Record` to `some:SpecificBibliographicNotionOfRecord` (subclassing `PlatormRecord`).

Alright! My only suggestion then is to document the definition/scope of this category in base.ttl

PR #416, opened 2023-02-21
Add binding property and class

klngwll:
Add  bf:binding and bf:Binding introduced in bf2.2 to kbv.

- [x] Semantic relations
- [x] Swedish label (needs to be verified)
- [ ] Swedish definition


niklasl (approved):
LGTM. Are there any valuable enumeration lists out there for reasonable values?

PR #419, opened 2023-02-24
Move translationOf position in card

klngwll:
Moving translationOf position in display cards to be more aligned with relations to other things (earlier placement because of it's relation to language and translated language).

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -601,7 +601,6 @@
             },
             "version",
             "marc:arrangementStatementForMusic",
-            "translationOf",
             "contribution",
             "language",
             "hasNotation",
@@ -615,7 +614,8 @@
             "illustrativeContent",
             "cartographicAttributes",
             "isPartOf",
-            {"inverseOf": "instanceOf"}
+            {"inverseOf": "instanceOf"},
+            "translationOf"
           ]
         },
         "DataCatalog": {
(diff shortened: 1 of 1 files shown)


source/vocab/display.jsonld
```diff
@@ -615,7 +614,8 @@
             "illustrativeContent",
             "cartographicAttributes",
             "isPartOf",
-            {"inverseOf": "instanceOf"}
+            {"inverseOf": "instanceOf"},
+            "translationOf"
```
olovy, line 618: Note that this will place {"inverseOf": "instanceOf"} in the "main form" instead of in the "incoming links" section.
Is that intended?

klngwll: [97217ab](https://github.com/libris/definitions/pull/419/commits/97217ab0a9e899f9b130a238aeb253124d8baf42)

olovy (approved):
:ship:

source/vocab/display.jsonld
```diff
@@ -615,7 +614,9 @@
             "illustrativeContent",
             "cartographicAttributes",
             "isPartOf",
-            {"inverseOf": "instanceOf"}
+            "translationOf"
+            {"inverseOf": "instanceOf"},
```
niklasl, line 618: No comma trailing comma,

klngwll: https://github.com/libris/definitions/commit/2c4417e87e15ab1a59fe8a70a86de4cf8df38e3c

PR #422, opened 2023-02-27
Overcome backwards xl-id form by handling both

niklasl:
This adds the "normal" form (used in XL) as a sameAs to generated records if they are to have the backwards form mistakenly minted by this repository.

An upper timestamp per dataset is used to check for if its member records are to have the backwards form.

(Eventually we want to "garbage collect" this backwards form from XL, to ensure they don't "squat" on XL id:s in the future.)


lxltools/datacompiler.py
```diff
@@ -189,10 +192,25 @@ def _compile_dataset(self, name, result):
                     node,
                     created_ms,
                     modified_ms,
-                    datasets=[self.dataset_id, ds_url])
-            self.write(desc, fpath)
+                    datasets=[self.dataset_id, ds_url],
+                    ds_created_ms=ds_created_ms)
+
+            # Keep sameAs "fowards" form in meta even if no_records is used
```
olovy, line 198:
```suggestion
            # Keep sameAs "forwards" form in meta even if no_records is used
```

PR #423, opened 2023-02-28
Add CC license metadata

niklasl:
- Define SPARQL construct for fixing problems in the metadata.
- Use Makefile for constructing "curated copies".
- Include note about license metadata licensing.
- Add some vocab term mappings to KBV.


olovy (approved):
:ship:

PR #425, opened 2023-03-15
Show romanization rule in language tag chips

olovy:
Display rule in language tag chips for transformed strings.
e.g. "Armenian · ALA-LC" instead of "Armenian · latin alphabet"


kwahlin (approved):
:ship:

PR #426, opened 2023-03-22
Update swedish label for instanceOf

klngwll:
Simplify Swedish label for instanceOf -> "instans av" instead of "instans av Verk".

Not needed for qualification when we get more titles in works.


olovy (approved):
LGTM

PR #427, opened 2023-03-22
Add Cast Class for validation

klngwll:
Add marc:Cast to vocab to for validation

`Unknown vocab value > key: marc:displayConstantController value: marc:Cast `


niklasl (approved):
If it exists it exists. :slightly_smiling_face: :shipit:

PR #428, opened 2023-03-31
Feature/add policies

niklasl:
Adding use policy for digital collections.

The crucial thing is to live with the URI and be OK for now with the (RDF) description. Then we can merge and iterate details when in QA.


andand79 (approved):
Looks good! If possible, add a colon to the acronym.
skos:prefLabel "KB:s villkor för digitala samlingar"@sv

niklasl:
@andand79 Good catch; added that in [5083e25](https://github.com/libris/definitions/pull/428/commits/5083e25abac4c82be42574ed60e31916cbb944cf).

PR #435, opened 2023-05-16
Add core KSSB classification components

niklasl:
This adds the KSSB (colloqially known as SAB)  core component terms as a dataset. These can be used to construct complex classification terms.

These are extracted by `extract_sab_data_from_docx.py` from the Word document source (right now we need to ask for a copy; [the published freely available source is a PDF](https://metadatabyran.kb.se/klassifikation/sab)). See the `Makefile` for the invocation.

Follow-up issues, if prioritized, would be to:
* Let the client search for and construct these (using `termComponentList`) as conveniently as possible.
* Extract existing constructs from string codes and link them.
* Link both simple component and constructed terms to DDC codes (when these are available as linked data). This would replace the ["conversion table"](https://metadatabyran.kb.se/klassifikation/ddk/mappning---konverteringstabell-ddk-sab).


klngwll (approved):
LGTM 🌟  I think we still need to be vary about the inScheme `version` coming back for a bite and maybe make a note of it somewhere. But for live-testing I don´t think we should complicate things further than having "sab-hubs" and that's where we need to get!

niklasl:
@klngwll Great! Merging this unless the `Subdivision`-subclasses may turn up in current cataloguing practices?

PR #437, opened 2023-05-17
Add image, thumbnail, ImageObject, contributor to vocab

olovy:
Add properties and classes for representing images of any kind of resource (e.g. instances/works/persons/concepts), and thumbnails thereof.

These are *specific* relationships, to overcome the vagueness and possibly contradictory interpretations of `associatedMedia` (and e.g. the *very* specific `hasRepresentation` (see also PREMIS disjointedness)).

### DIFF
--- a/source/vocab/relations.ttl
+++ b/source/vocab/relations.ttl
@@ -195,7 +195,15 @@
     rdfs:domain :Resource;
     rdfs:range :ImageObject;
     rdfs:subPropertyOf :isPrimaryTopicOf;
-    owl:equivalentProperty sdo:image .
+    owl:equivalentProperty sdo:image, foaf:depicts .
+
+ :imageOf a owl:ObjectProperty;
+    rdfs:label "image of"@en, "bild på"@sv;
+    rdfs:comment "Det som finns på bilden."@sv;
+    rdfs:domain :ImageObject;
+    rdfs:range :Resource;
+    rdfs:subPropertyOf sdo:about ;
+    owl:equivalentProperty foaf:depicts .    
 
 :thumbnail a owl:ObjectProperty;
     rdfs:label "thumbnail"@en, "miniatyrbild"@sv;
--- a/source/vocab/relations.ttl
+++ b/source/vocab/relations.ttl
@@ -191,15 +191,15 @@
 
 :image a owl:ObjectProperty;
     rdfs:label "image"@en, "bild"@sv;
-    rdfs:comment "En bild på resursen"@sv;
+    rdfs:comment "En bild på resursen."@sv;
     rdfs:domain :Resource;
     rdfs:range :ImageObject;
     rdfs:subPropertyOf :isPrimaryTopicOf;
     owl:equivalentProperty sdo:image, foaf:depicts .
 
- :imageOf a owl:ObjectProperty;
+:imageOf a owl:ObjectProperty;
     rdfs:label "image of"@en, "bild på"@sv;
-    rdfs:comment "Det som finns på bilden."@sv;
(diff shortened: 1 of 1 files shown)


source/vocab/relations.ttl
```diff
@@ -594,6 +616,14 @@
     ptg:abstract true;
     owl:equivalentProperty bf2:contributionOf .
 
+:contributor a owl:DatatypeProperty;
+    :category :shorthand;
+    rdfs:label "Contributor"@en, "Medverkande"@sv;
+    rdfs:domain :Document; # TODO?
```
niklasl, line 622:
```suggestion
    sdo:domainIncludes :Endeavour ;
```
Same as for `:contribution` (which admittedly isn't ideal, but works for now).

source/vocab/relations.ttl
```diff
@@ -189,6 +189,28 @@
     rdfs:subClassOf :Document;
     owl:equivalentClass sdo:MediaObject .
 
+:image a owl:ObjectProperty;
+    rdfs:label "image"@en, "bild"@sv;
+    rdfs:comment "En bild på resursen"@sv;
+    rdfs:domain :Resource;
+    rdfs:range :ImageObject;
+    rdfs:subPropertyOf :associatedMedia;
```
niklasl, line 197:
```suggestion
    rdfs:subPropertyOf :isPrimaryTopicOf;
```
Since `associatedMedia` has a narrower domain (`:Endeavour`), we can't use it as a base property here. `isPrimaryTopicOf` works (though I'm not sure how useful it is; perhaps just skip any base for this).

niklasl (changes requested):
Looks good! Some small things to address, but in general good additions to have.

source/vocab/relations.ttl
olovy: But then it doesn't work on `ImageObject`?

olovy:
```suggestion
    rdfs:domain :Resource;
```
We want to be able to add an image on any Resource?

source/vocab/relations.ttl
```diff
@@ -189,6 +189,28 @@
...
+    rdfs:subPropertyOf :isPrimaryTopicOf;
+    owl:equivalentProperty sdo:image .
+
+:thumbnail a owl:ObjectProperty;
+    rdfs:label "thumbnail"@en, "miniatyrbild"@sv;
+    skos:altLabel "tumnagel"@sv, "indexbild"@sv;
+    rdfs:comment "En förminskad bild"@sv;
+    sdo:domainIncludes :ImageObject;
+    sdo:rangeIncludes :ImageObject;
+    rdfs:subPropertyOf :image;
```
niklasl, line 206: Is this *between these* to relationships? Neither just an image of the resource, nor just another format thereof; but both.

```suggestion
    rdfs:subPropertyOf :image, :hasFormat ;
```

niklasl (commented):
Some more suggestions.

source/vocab/relations.ttl
niklasl: ... But that (`hasFormat`) doesn't hold for a thumbnail of a video! `hasVersion` sort of works. I'm fine if we reject this for now. (Would be good to note this fact in a comment in the ttl though.)

source/vocab/relations.ttl
olovy:
```suggestion
    sdo:domainIncludes :Endeavour, :MediaObject ;
```

niklasl (approved):
I think this is good to go! :shipit:

PR #441, opened 2023-06-19
Update domainIncludes for subject to exclude Instance

klngwll:
Prohibits subjects to be added to the Instance descriptions.

Follows up LXL-3167


niklasl (approved):
LGTM!

(Looking forward to be able to also exclude `:subject` on `:Item` ...)

PR #443, opened 2023-08-21
Update translationOf in work card order

klngwll:
Move translationOf higher in order of cards (not full view because of workflow differences) to ease spotting the correct work.

See LXL-4193


kwahlin (approved):
LGTM

livdam (commented):
LGTM

PR #445, opened 2023-09-18
Add labels for Endeavour, Creation and Embodiment

niklasl:
The English labels are just the formal name, the Swedish ones are in the "X or Y" form suggested. Use these for the English ones too?

(Note: `Embodiment` is also the baseclass of `Representation` (which is from PREMIS and has an "interesting" relationship to `bf:Item`)...)


kwahlin (approved):
Looks good. Regarding the English labels...a coin flip to me. Both ways make sense.

livdam (approved):

katarinasynnermark (approved):

PR #446, opened 2023-09-27
move illustrativeContent to Instance card

klngwll:
Move `illustrativeContent` from Work to Instance Card.

Suggestion just below `PhysicalDetailsNote` which also holds illustrative information.


kwahlin (approved):
:shipit:

PR #452, opened 2023-11-01
Add standalone Work-full lens

klngwll:
Removing Work full lens-extension from Card. To not chase our tails so we can both have a relation to translationOf as a distinct related entity last in the fullview and having an original title and language readily available in the card view through `translationOf/hasTitle|language` .


kwahlin (approved):
LGTM!

PR #453, opened 2023-11-08
Add classes and properties for "CXZ"

olovy:
The naming of these concepts can be improved but i believe the structure is working for now.

Skärmbild från 2023-11-08 18-34-27


lrosenstrom (approved):
Nice, lgtm!

PR #462, opened 2024-01-18
Add fresnel:Format definitions

olovy:
WIP draft

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -103,7 +103,7 @@
           "@id": "Resource-chips",
           "@type": "fresnel:Lens",
           "classLensDomain": "Resource",
-          "showProperties": [ 
+          "showProperties": [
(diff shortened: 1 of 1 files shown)


source/vocab/display.jsonld
```diff
@@ -1436,14 +1435,186 @@
...
+        "date",
+        "marc:publicationStatus"
+      ],
+      "fresnel:propertyFormat": {
+        "fresnel:contentBefore": ", ",
+        "fresnel:contentFirst": ""
+      }
+    },
+    "ProvisionActivity-endYear-format": {
+      "TODO": "endYear without startYear? fresnel:contentNoValue on startYear? but we only want it when endYear is present",
```
olovy, line 1513: And the opposite of course.
We need a way to create virtual properties in showProperties or something.
(this is not what fresnel:mergeProperties does. It just merges the values of multiple properties to a single one)

PR #464, opened 2024-01-22
Add identifier RAiD as owl:Class in details.ttl

andand79:


niklasl (commented):
Looks good! Could you also add some definition reference (see uses of `rdfs:seeAlso` and `skos` matches on other `Identifier` subclasses)?

(As a follow-up we may go trough the existing ones and perhaps change some chosen relationships (e.g. to `dct:references`).)

andand79:
With the identifier being new, no other exactMatch have been found. subClassOf changed to URI since the identifiers are said to be expressed as actionable URLs.

source/vocab/details.ttl
```diff
@@ -694,7 +694,8 @@
     :category :pending ;
     rdfs:label "RAiD";
     rdfs:comment "Research Activity Identifier";
-    rdfs:subClassOf :Identifier .
+    skos:definition "Persistent identifier for research projects; governed by ISO 23527 and managed by Australian Research Data Commons."@en;
```
niklasl, line 697: Add e.g.:
```
    rdfs:seeAlso <https://raid.org/> ;
```
if that's the official site for its definition.

niklasl (approved):
LGTM!

andersju:
I assume we can go ahead and merge this? :)

andand79:
Yes! At the moment I'm not authorised to merge so please go ahead.

PR #465, opened 2024-01-23
Swepub: add PublicationStatus Retracted

andersju:
https://jira.kb.se/browse/SWEPUB2-1112

Should we perhaps at the same time remove the seemingly unused `swepub:Redacted`?


andand79 (approved):
Yes, swepub:Redacted was incorrectly created two months ago and has not been communicated, so it should be straightforward to remove it.

PR #466, opened 2024-01-29
Additions and changes  (mostly) related to Swepub and auto-classification

niklasl:
WIP (to be discussed):

* [x] Add SoftwareAgent class (To differentiate between a process and its used software or service. (GenerationProcess is a conflation between the two; and a subclass of AdminMetadata. We may want to improve this across the platform.))
* [x] Update SSIF generation script.


niklasl:
I intend to merge this today, to work on it in QA. It does not need to move into production yet, as QA is reasonably enough to take out and cache the data Swepub needs: for A) auto-classification, and B) the SSIF tree in its search UI.

We will use this external source ("fake" until published through UKÄ) by mapping it into KBV and build on that integration for the Swepub needs (which follows alignment of Swepub with Libris through KBV).

klngwll (approved):
LGTM!

PR #474, opened 2024-03-01
Add hasComponent to Item search-chip

olovy:
So that the Items in hasComponent can be searched in the same way as the top-level Item.

Example: shelfMark in @reverse.itemOf
`/find?%40type=Instance&@reverse.itemOf.hasComponent.shelfMark.label=KoB*` 
should work in the same way as
`/find?%40type=Instance&@reverse.itemOf.shelfMark.label=KoB*`


kwahlin (approved):
LGTM

PR #477, opened 2024-03-14
Remove archival terms

klngwll:
Removal of archival terms from BIB 000/06[b] (0 uses and obsolete since 1995) and 000/08 (noisy info on description purpose).

- [x] [librisxl PR](https://github.com/libris/librisxl/pull/1388) with Marcframe changes and Cleanupscript for marc:typeOfControl.

FMT-229


niklasl (approved):
:shipit:

PR #481, opened 2024-04-03
Add filter on concerning.issuanceType

olovy:
Depends on https://github.com/libris/librisxl/commit/2e7a15a3d4957193251013fdb4163feda43cf681

Maybe we should add a mechanism to the search API that checks which keyword mappings are available before enabling a filter.
Otherwise the upgrade procedure is more involved. The index needs to be recreated before the filter can be added to apps.jsonld.


kwahlin (approved):
LGTM!

PR #487, opened 2024-05-27
Change linked description to linked articleList in apps.jsonld

niklasl:
Addresses indexing problem. We should also define `article` and `articleList` for these.


olovy (approved):
:ship:

PR #492, opened 2024-07-31
Add support for manual transliteration of Iranian languages (other)

olovy:
They mostly use ALA-LC romanization schemes. For other cases we leave the transliteration scheme unspecified.
(This results in two input fields in the cataloging interface. The one not used can be left empty.)


andersju (approved):
LGTM!

PR #494, opened 2024-08-19
Add SingleItem as subclass to Item

olovy:
This is a step on the way to https://github.com/libris/definitions/pull/483

### DIFF
--- a/source/vocab/items.ttl
+++ b/source/vocab/items.ttl
@@ -33,6 +33,12 @@
     owl:equivalentClass bf2:Item, sdo:Product, holding:Item .
     #TODO?: owl:equivalentClass [ owl:unionOf (sdo:Product sdo:Offer) ] ;
 
+:SingleItem a owl:Class ;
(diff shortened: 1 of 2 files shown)


source/vocab/items.ttl
```diff
@@ -210,7 +216,7 @@
 :itemUsed a owl:ObjectProperty ;
     rdfs:label "Item used"@en, "Exemplaret som använts"@sv ;
     rdfs:domain :ProvisionActivity ;
-    rdfs:range :Item .
+    rdfs:range :SingleItem .
```
niklasl, line 219: IIUC, sometimes it is unknown witch exact item was used; so this should probably stay as just `:Item`? Ping @ebengtsson

olovy: Sounds reasonable

olovy, line 219:
```suggestion
    rdfs:range :Item .
```

PR #497, opened 2024-09-20
Update definition of :maxItems

olovy:
It is now also used to signal the max number of items in a Slice (facet group). 
Corresponding to `itemLimit` in apps.jsonld. Which is not defined and could be replaced with `:maxItems`?


niklasl:
Yes, replacing itemLimit with maxItems in apps.jsonld sounds reasonable.

PR #500, opened 2024-10-17
Fix series relationsships domain and range

klngwll:
Fixes incorrect domains and range for some relationships used for series.

Marcframe conversion:
```
780 "aboutEntity": "?thing",
  {"when": "i2=3", "addLink": "precededInPartBy", "resourceType": "Instance"},
  {"when": "i2=6", "addLink": "absorbedInPart", "resourceType": "Instance"},
785 "aboutEntity": "?thing",
  {"when": "i2=3", "addLink": "succeededInPartBy", "resourceType": "Instance"},
  {"when": "i2=5", "addLink": "absorbedInPartBy", "resourceType": "Instance"},

```

Reference FMT-42

### DIFF
--- a/source/vocab/relations.ttl
+++ b/source/vocab/relations.ttl
@@ -105,8 +105,8 @@
 :absorbedInPart a owl:ObjectProperty;
     rdfs:label "Absorbed in part"@en, "Har delvis införlivat"@sv;
     rdfs:subPropertyOf :precededBy;
-    rdfs:domain :Work;
-    rdfs:range :Work;
+    rdfs:domain :Instance;
+    rdfs:range :Instance;
     rdfs:comment "Work that has been partially incorporated into another work."@en;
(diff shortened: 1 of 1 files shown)


niklasl:
We are looking at incorporating [Serial as a subclass of work](https://id.loc.gov/ontologies/bibframe.html#c_Serial), as per BF 2.3(?). Shouldn't we change marcframe instead; or should we "worsen" this first, and move it all to the right shape for all serials?

klngwll:
> We are looking at incorporating [Serial as a subclass of work](https://id.loc.gov/ontologies/bibframe.html#c_Serial), as per BF 2.3(?). Shouldn't we change marcframe instead; or should we "worsen" this first, and move it all to the right shape for all serials?

As discussed, updated with `:Creation` to not close any doors for future development while still being useful now.

niklasl (approved):
LGTM :shipit:

PR #501, opened 2024-10-25
Establish bulk: namespace

kwahlin:


source/vocab/platform.ttl
```diff
@@ -28,6 +28,7 @@
 
 @prefix : <https://id.kb.se/vocab/> .
 @prefix marc: <https://id.kb.se/marc/> .
+@prefix bulk: <https://id.kb.se/bulk/> .
```
niklasl, line 31: Maybe:
```suggestion
@prefix bulk: <https://id.kb.se/ns/bulk/> .
```
to make a (name)space for "other vocabs".

(Could also be under `/vocab/bulk/` ,but that'd make it appear subordinated to KBV rather than orthogonal. The reason for `/ns/` is also to establish a pattern for ontology terms as separate from enumeration lists. These often go under `/term/` _if_ they belong to a concept scheme, but some are directly in the "root".)

sys/context/kbv.jsonld
```diff
@@ -9,10 +9,10 @@
     "recordStatus": {"@type": "@vocab"},
     "encodingLevel": {"@type": "@vocab"},
     "shelfMarkStatus": {"@type": "@vocab"},
-    "bulkChangeStatus": {"@type": "@vocab"},
-    "bulkChangeMetaChanges": {"@type": "@vocab"},
     "dimensionChain": {"@id": "dimension", "@type": "@vocab", "@container": "@list"},
     "inverseOfTerm": {"@id": "owl:inverseOf", "@type": "@vocab"},
+    "bulk:status": {"@type": "@vocab"},
```
niklasl, line 14: We also need the `bulk` prefix declared. Also, a weakness in JSON-LD is that it "blocks" a term called `bulk`. (Though in JSON-LD 1.1 it can be namespace-only; but that's mainly for avoiding expanding prefixes that are regular terms; it doesn't allow a term to share the name with a prefix.)

PR #502, opened 2024-10-31
Add bulk merge terms

kwahlin:
Or is it rather a `bulk:Deprecation` at this point? Until we actually merge data from different records.


olovy:
Let's got with Merge

olovy:
> Or is it rather a `bulk:Deprecation` at this point? Until we actually merge data from different records.

Let's go with Merge

olovy (approved):
LGTM! 
Merge Merge!

PR #506, opened 2024-11-19
Add :Identity to domain of :meta

kwahlin:
Enables adding `:meta` property to `:Agent` and `:Concept` resources (needed in the bulk change tool).

Should make sense since `:Identity` resources often appear as `:mainEntity` and we have `:meta owl:inverseOf :mainEntity`.


niklasl (approved):
In principle, anything can be a mainEntity of a record, so it makes sense logically to have a resource domain (which really is implied). The crux is to ensure that you cannot add these properties unless "we find it useful".

I doubt that anyone accidentally adds these on inststances of Identity subclasses though; so I think this is OK here. But this is important to keep in mind in general. If demands for what should be forbidden in the cataloging tool are at odds with that is logically true, we need to revisit how to control what is allowed to be added in the cataloging tool (including using SHACL to control this explicitly).

olovy (approved):

PR #512, opened 2025-02-07
Ensure lens index key and classLensDomain are the same

niklasl:
The indexed key seems to be correct and the one actually used. The actual `classLensDomain` is (worryingly) wrong.

(This is a recurring problem best addressed by removing the well-intentioned but complicating reliance on `@container: @index` in this file. I'd gladly explain this in more detail if there is time.)

### DIFF
--- a/scripts/checkdisplay.py
+++ b/scripts/checkdisplay.py
@@ -1,0 +1,20 @@
+import json
+from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


branchedelac (approved):
Ser bra ut!

olovy (approved):
LGTM

PR #514, opened 2025-02-17
Add RDA DescriptionConventions to kbv-enums

olovy:
Add entity for RDA description conventions / cataloging rules.

See https://kbse.atlassian.net/browse/LXL-3651


kwahlin (approved):
LGTM! Could add `:DescriptionConventions :category :heuristicIdentity` to trigger automatic linking by the XL normalizer. If there is a need, not sure if these markers are ever set elsewhere than in the cat client.

olovy:
> LGTM! Could add `:DescriptionConventions :category :heuristicIdentity` to trigger automatic linking by the XL normalizer. If there is a need, not sure if these markers are ever set elsewhere than in the cat client.

Good idea. I'll do that.

PR #525, opened 2025-04-29
Add filter for free digital material

lrosenstrom:
See description in [LWS-242](https://kbse.atlassian.net/browse/LWS-242). 

* Add optional filter for free e-resources for lxl-web/Libris sök. The filter can be formulated in the lxl web query language like so (which is sort of a hack, but will do until the free nature of things is expressed more clearly in the data): 
```
(usageAndAccessPolicy.label:"gratis" OR "associatedMedia.marc:publicNote":"*fritt tillgänglig*") AND hasInstanceType:"Electronic"
```
 * Add MediaObject-cards definition to include marc:publicNote in the search index, and hence facilitate the query above.

Edit: @olovy pointed out that we should probably add https://id.kb.se/policy/freely-available and https://id.kb.se/policy/oa/gratis to the filter, so those two access policies are also included now.

### DIFF
--- a/source/apps.jsonld
+++ b/source/apps.jsonld
@@ -98,7 +98,7 @@
         { "alias": "excludePreliminary", "filter": "NOT encodingLevel:(\"marc:PartialPreliminaryLevel\" OR \"marc:PrepublicationLevel\")", "prefLabelByLang": { "sv": "Exkludera kommande publiceringar", "en": "Exclude upcoming publications" } },
         { "alias": "includePreliminary", "filter": "NOT excludePreliminary", "prefLabelByLang": { "sv": "Inkludera kommande publiceringar", "en": "Include upcoming publications" } },
         { "alias": "existsImage", "filter": "image:*", "prefLabelByLang": { "sv": "Endast resurser med omslags-/miniatyrbild", "en": "Resources with cover/thumbnail only" } },
-        { "alias": "freeDigital", "filter": "hasInstanceType:\"Electronic\" AND (usageAndAccessPolicy.label:\"gratis\" OR \"associatedMedia.marc:publicNote\":\"*fritt tillgänglig*\")", "prefLabelByLang": { "sv": "Fritt digitalt", "en": "Free digital material" } }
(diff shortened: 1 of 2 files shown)


source/apps.jsonld
```diff
@@ -97,7 +97,8 @@
...
         { "alias": "excludePreliminary", "filter": "NOT encodingLevel:(\"marc:PartialPreliminaryLevel\" OR \"marc:PrepublicationLevel\")", "prefLabelByLang": { "sv": "Exkludera kommande publiceringar", "en": "Exclude upcoming publications" } },
         { "alias": "includePreliminary", "filter": "NOT excludePreliminary", "prefLabelByLang": { "sv": "Inkludera kommande publiceringar", "en": "Include upcoming publications" } },
-        { "alias": "existsImage", "filter": "image:*", "prefLabelByLang": { "sv": "Endast resurser med omslags-/miniatyrbild", "en": "Resources with cover/thumbnail only" } }
+        { "alias": "existsImage", "filter": "image:*", "prefLabelByLang": { "sv": "Endast resurser med omslags-/miniatyrbild", "en": "Resources with cover/thumbnail only" } },
+        { "alias": "freeDigital", "filter": "hasInstanceType:\"Electronic\" AND (usageAndAccessPolicy.label:\"gratis\" OR \"associatedMedia.marc:publicNote\":\"*fritt tillgänglig*\")", "prefLabelByLang": { "sv": "Fritt digitalt", "en": "Free digital material" } }
```
olovy, line 101: `"fritt tillgänglig\"` (without the asterisks) should be sufficient?

olovy (commented):
Note that we can't be sure that e.g. usageAndAccessPolicy and type:Electronic match the same instance since instances are not nested in ES. This should not be a problem in practice. And as you say, the filter should be based on more exact symbols.

See comment about wildcards. Other than that, LGTM

source/apps.jsonld
kwahlin: Also no need for quotation marks around `Electronic` and `gratis`!

olovy (approved):

kwahlin (approved):
LGTM

PR #533, opened 2025-07-03
Add uv project

andersju:
See https://kbse.atlassian.net/browse/LXL-4675

I removed `.python-version` from `.gitignore` because consensus seems to be that it _should_ be checked in, to make sure we're all running the same version. (`uv init foo` creates both `foo/.gitignore` and `foo/.python-version` and doesn't git-ignore the latter.) From one of the main uv authors:

> The information [`.python-version`] captures is materially different than `requires-python`. `requires-python` just tells you the range of supported versions; meanwhile, teams use `.python-version` to dictate the version that should be used during development. That’s useful both for libraries and applications.
>
> It also does not need to be a patch version – you can put `3.12` or similar to ensure that everyone is using Python 3.12, rather than anything later or earlier.

(https://discuss.python.org/t/standardize-python-version-file-in-pyproject-toml/86342/18)

Keeping `requirements.txt` around for now but at some point it should be removed.


olovy (approved):
It works!
LGTM.

See comment about version in pyproject.toml

PR #535, opened 2025-08-26
Add tableOfContents to Instance search-card, remove unnecessary fsl selectors from Work search-card

kwahlin:
Depends on https://github.com/libris/librisxl/pull/1619


olovy (approved):
:ship:

PR #536, opened 2025-09-10
Map :controlNumber to ONR query code

kwahlin:
Example usage with in new XL find API: `/find?_q=ONR:13576674`

`ONR` maps to `:controlNumber` and since `:controlNumber` has the domain `:Record`, it automatically expands to `meta.controlNumber` (see https://github.com/libris/librisxl/blob/a7cbe4d8ffe7a05a484d9b5c314c4be08ccb2a30/whelk-core/src/main/groovy/whelk/search2/querytree/Path.java#L111).


andersju (approved):
Nice, thanks! LGTM!

PR #576, opened 2026-03-31
Feature/remodel instance vs work search

kwahlin:
- Control instance vs. work search term behavior via domain/range rather than via property chain axioms (makes much more sense in my opinion). Whether `instanceOf`/`hasInstance` should be prepended in search can be inferred from the domain.
- Make `ls:workCategory` a composite property with `ls:findCategory`, `ls:identifyCategory`, and `ls:noneCategory` as sub-properties. Previously, values outside `ls:findCategory` and `ls:identifyCategory` defaulted to `ls:noneCategory`. However, some use cases (e.g. free-text search within workCategory) require querying across all three.
- Demonstrate behavior for `ls:findCategory `/`ls:identifyCategory`/`ls:noneCategory`  by setting `:category :coercing` (should be straight-forward to add corresponding sub-properties of `ls:instanceCategory` if/when needed)
- Show workCategory and instanceCategory as separate incoming relations (format/kategori): [77061eb](https://github.com/libris/definitions/pull/576/commits/77061eb887fe7f3031c311e5d46141515f766fa4)
- Move `rdf:type` to the top of `sliceList` so it will be overridden by `workType`/`instanceType` when those are applicable. (This doesn't work for mixed-type queries, e.g. `type:Agent OR type:Instance`, however mixed-type querying is already broken for other reasons so let's address that later.)
- Related changes in librisxl: https://github.com/libris/librisxl/pull/1736
- Related changes in lxlviewer: https://github.com/libris/lxlviewer/pull/1541 (already merged and then reverted ([cd741cc](https://github.com/libris/lxlviewer/commit/cd741ccbf4b9dded467390ff580b315a0cf57da5), needs re-revert)
- New API tests include these terms: ~https://github.com/libris/lxl_api_tests/pull/44 (tests only basic querying so e.g. aggregating needs manual testing)~ https://github.com/libris/lxl_api_tests/pull/45


source/apps.jsonld
```diff
@@ -122,7 +122,7 @@
       ],
       "relationFilters": [
         { "objectType": "Agent", "predicates": [ "contributor", "subject", "publisher" ] },
-        { "objectType": "Concept", "predicates": [ "subject", "genreForm", "category", "hasOccupation", "fieldOfActivity", "termComponentList" ] },
+        { "objectType": "Concept", "predicates": [ "subject", "genreForm", "workCategory", "instanceCategory", "fieldOfActivity", "termComponentList" ] },
```
olovy:
```suggestion
        { "objectType": "Concept", "predicates": [ "subject", "genreForm", "workCategory", "instanceCategory", "hasOccupation", "fieldOfActivity", "termComponentList" ] },
```

olovy (approved):
LGTM!

PR #580, opened 2026-04-17
Feature/item search filter

kwahlin:
Enables using `bestånd:` for searching:
- "hyllsignum"
- "lokal anmärkning"
- "lokalt ämnesord"

**TODO?** 
Align search filters with how data is presented in the item panel:
- Replace`ls:hasItemNote` (lokal anmärkning) with a filter that matches values shown under Beståndsuppgift
- Replace `ls:itemShelf` (hyllsignum) with a filter that matches values shown under Placering
- Add an additional filter to cover remaining fields to be searchable via `bestånd:`


olovy (approved):
LGTM!

PR #581, opened 2026-04-21
Feature/bibliography broader

olovy:
* Add descriptionHTML to context. To be used for bibliography descriptions
* Add parts to bibliographies BOTH and OSGO
* Define "virtual" bibliographies SVETRYCK and NATBIB
    * `:category div:abstract` works for blocking in cataloging but is meant for vocab terms?


niklasl (approved):
LGTM :shipit:

niklasl:
Do we want to add `div:deprecated` in cataloguing? It would be more generally applicable than `div:abstract`.

olovy:
> Do we want to add `div:deprecated` in cataloguing? It would be more generally applicable than `div:abstract`.

Maybe. But these bibliographies are not exactly deprecated. More virtual / abstract / grouping

PR #583, opened 2026-04-24
Add composite search filter ls:library

kwahlin:
Refactor how libraries are queried:

- Downgrade `ls:itemHeldBy` and `ls:itemHeldByOrg` to auxiliary sub-properties.
- Introduce `ls:library` as a composite super-property, now used as the primary filter for searching both `Library` and `bibdb:Organization`.
- Mark the sub-properties with `:category :coercing` to mirror the behavior of `ls:workCategory` and its sub-properties.
- Retain `ls:itemHeldByOrg` in facets where results should be limited to organizations.
- Adjustments in XL: https://github.com/libris/librisxl/pull/1754  
- Minor adjustment in API tests: https://github.com/libris/lxl_api_tests/pull/47


olovy (approved):
LGTM!

PR #585, opened 2026-04-27
Refine item search filters

kwahlin:
In accordance with what has been documented in Excel files. Thoughts on property naming and labels? 

See also https://github.com/libris/librisxl/pull/1759

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -1765,6 +1765,7 @@
                 "shelfLabel",
                 "physicalLocation",
                 "shelfControlNumber",
+                "copyNumber",
                 {
                   "@type": "fresnel:fslselector",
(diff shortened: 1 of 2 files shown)


source/vocab/display.jsonld
```diff
@@ -1702,11 +1701,9 @@
           "showProperties": [
             "fresnel:super",
             "subject",
-            "summary",
             "hasComponent",
             {
               "fresnel:mergeProperties": [
-                "cataloguersNote",
                 "custodialHistory",
```
olovy: custodialHistory should only be in additionalItemInformation?

kwahlin: Yes!

olovy (approved):
LGTM!

PR #586, opened 2026-04-27
Add term definitions for uppsok

niklasl:
Define terms for concepts used in Uppsök; specifically:
* [x] Work categories
* [x] Classifications
* [x] new scheme

### DIFF
--- a/lxltools/datacompiler.py
+++ b/lxltools/datacompiler.py
@@ -115,10 +115,11 @@ class Compiler:
     def path(self, pth):
         return self.base_dir / pth
 
-    def to_jsonld(self, graph):
+    def to_jsonld(self, graph, context=None):
+        context = (context or self.context)
         return ldutil.to_jsonld(graph,
-                         "../" + self.context,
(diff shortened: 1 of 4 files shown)


kaipoykio (commented):
LGTM ;)

branchedelac (approved):
Looks good to me!

kaipoykio (changes requested):
Is it possible to change the setname classification names to corresponding 'Set Spec' identifiers instead (https://uppsok.libris.kb.se/websearchhelp/help_uppsok/metadatamodell.htm#uppsa) ? It would remove one level of translation in the conversion code ;)

niklasl:
Makes sense to use those ("set spec") names for the local name parts! (IIRC we made just them up and missed those formal precursors.)

niklasl:
Revised in 8b97957e17cd0ee0ca9c042e21fc309faea61291 to use set spec names.

kaipoykio (approved):
Nice!

PR #591, opened 2026-05-08
Add terms used for Shared Print Actions (FMT-214)

niklasl:
Define classes, properties and concept terms used for describing Shared Print Actions. (FMT-214)


andand79 (approved):
LGTM

PR #593, opened 2026-05-12
Add MediaObject search chip including marc:publicNote

kwahlin:
To make `marc:publicNote` searchable in Libris


olovy (approved):
LGTM

I wonder if we should do isPrimaryTopicOf/Document as well
https://api.triplydb.com/s/tH_v-Ks8U

PR #596, opened 2026-05-21
Feature/instance record search filters

kwahlin:
- Fix definitions to get correct query expansion for `DAT` / `BDAT` query codes.
- Ensure `meta.created`, `meta.modified` and `meta.encodingLevel` are indexed for embedded instances.

Maybe we should create similar shorthands for all record search filters? For example define
```
ls:bibliography owl:propertyChainAxiom ( [ rdfs:subPropertyOf :meta ; rdfs:domain :Instance ] :bibliography ) .
```
I think this would simplify things because then the correct query expansion can be inferred directly from the definition instead of having special handling of record properties in the implementation.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -1865,6 +1865,18 @@
             {
               "@type": "fresnel:fslselector",
               "@value": "meta/*/controlNumber"
+            },
+            {
+              "@type": "fresnel:fslselector",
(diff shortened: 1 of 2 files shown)


olovy:
> Maybe we should create similar shorthands for all record search filters?

Agreed. but not all record filters are locked to instance or work?

olovy (approved):
LGTM!

PR #598, opened 2026-05-21
Make additionalClassificationDdc available for search

kwahlin:
Add search filter corresponding to legacy code `DDC2` as well as a composite filter for searching both `DDC` and `DDC2`.

### DIFF
--- a/source/vocab/display.jsonld
+++ b/source/vocab/display.jsonld
@@ -1899,8 +1899,17 @@
             "relationship",
             "hasPart",
             {
-              "@type": "fresnel:fslselector",
-              "@value": "additionalClassificationDdc/ClassificationDdc/code"
+              "fresnel:mergeProperties": [
+                {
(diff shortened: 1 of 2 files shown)


olovy (changes requested):
There should only be one filter dewey/DDC/DDK which searches both classification and addititionalClassificationDdc

kwahlin:
> There should only be one filter dewey/DDC/DDK which searches both classification and addititionalClassificationDdc

See new solution: [b0f8f38](https://github.com/libris/definitions/pull/598/commits/b0f8f3837acba5a56309a339a1c5296bf1ff5ad6)

We can remove [this row](https://github.com/libris/librisxl/blob/1eead4b38e805938128e4c34694b75a502123c49/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy#L740) too after merging this.

olovy (approved):
LGTM!

PR #600, opened 2026-05-29
Feature/libris search filter adjustments

kwahlin:
Then main features here are:
- Widened scope for `ls:title` (now includes `hasPart.hasTitle`, `hasPart.translationOf.hasTitle`)
- New terms `ls:contributor`/`ls:author`/`ls:translator` (to include  `hasPart.contribution` and `relationship.contribution`)
- New term `ls:linkisxn` for searching `marc:incorrectIssn` / `marc:canceledIssn` / `:indirectlyIdentifiedBy` in addition to `isxn`
- Adjusted `excludePreliminary` filter to always allow linked works through.
    - These were previously filtered out incorrectly: https://api.triplydb.com/s/iEhP8sk61
    - This case will now pass the filter but that seems acceptable :slightly_smiling_face: 
    https://api.triplydb.com/s/hr8KrnhMB

Also includes:
- Record search terms defined in the `librissearch:` namespace for better control over query expansion without having to mess with their `kbv:` counterparts.
- Restructuring of the `.ttl` file
- Minor fixes discovered while testing

Related changes:
- https://github.com/libris/lxl_api_tests/pull/50
- https://github.com/libris/librisxl/pull/1780
- https://github.com/libris/lxlviewer/tree/feature/locales

TODO/Questions (nothing urgent, just noted things):
- Should `indirectyIdentifiedBy` be included directly in `:isbn`? (and `marc:canceledIssn` / `marc:incorrectIssn`  in `:issn`). I think that would make sense, although it would mean `:isxn` == `:linkisixn`, which differs from webbsök.  
- Index `marc:canceledIssn` / `marc:incorrectIssn` as keyword in Elastic
- For uniformity, would it be better to define composite search filters for e.g. `seriesMembership` rather than relying on `_str`? That way we can display exactly which fields are being searched.

### DIFF
--- a/source/vocab/libris-search-experimental.ttl
+++ b/source/vocab/libris-search-experimental.ttl
@@ -175,6 +175,10 @@ ls:encodingLevel a owl:ObjectProperty;
     rdfs:range marc:EncLevelType ;
     owl:propertyChainAxiom ( [ rdfs:subPropertyOf :meta ; rdfs:domain :Instance ] :encodingLevel ) .
 
+ls:recordType a owl:ObjectProperty;
+    :category :shorthand, :pending ;
+    owl:propertyChainAxiom ( :meta rdf:type ) .
+
 ls:isxn a owl:DatatypeProperty ;
     :category ls:composite, :pending, :searchfilter ;
     skos:notation "NUMM"^^ls:QueryCode ;
--- a/source/vocab/libris-search-experimental.ttl
+++ b/source/vocab/libris-search-experimental.ttl
@@ -407,7 +407,6 @@ ls:lccn rdfs:domain :Instance ;
     :category :searchfilter, :shorthand, :pending ;
     rdfs:label "LCCN";
     skos:notation "bath.lccn"^^ls:QueryCode ;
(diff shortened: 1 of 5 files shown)


source/vocab/libris-search-experimental.ttl
```diff
@@ -231,17 +274,212 @@ ls:noneCategory a :ObjectProperty ;
...
+
+ls:instanceCategory a owl:ObjectProperty ;
+    :category :searchfilter, :impliedByObject, :pending ;
+    rdfs:label "format"@sv, "format"@en ;
+    rdfs:domain :Instance ;
+    rdfs:range :InstanceCategory ;
+    ls:indexKey "_categoryByCollection.@none" .
+
+ls:linkisxn a owl:DatatypeProperty ;
+    :category ls:composite, :pending, :searchfilter ;
```
olovy:
```suggestion
    :category ls:composite, :pending ;
```

olovy (approved):
LGTM.

Let's wait with making LINKISXN visible.
See suggestion

PR #603, opened 2026-06-04
Make relation|relationship mappings explicit

niklasl:
These were based on BFLC. Their eventual incorporation swapped the meaning of relation and relationship (both properties and classes). We keep the former meaning since we're in production with those.

This requires suppressing automatic mappings to BF.

(Note that the automatic mapping process needs to be replaced by a controlled update routine; but that's a separate change.)

### DIFF
--- a/source/vocab/bf-map.ttl
+++ b/source/vocab/bf-map.ttl
@@ -46,6 +46,12 @@ kbv:relatedTo owl:equivalentProperty bf2:relatedTo ;
     rdfs:domain kbv:Resource ; # NOTE: loosens implied bf2 domain
     rdfs:range kbv:Resource . # NOTE: loosens implied bf2 range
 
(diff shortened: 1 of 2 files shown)


livdam (approved):

andand79 (approved):
Agreed

livdam (commented):
Should we also have a different swedish label for one of the properties and classes?

katnai1 (approved):
Agree with Liv to differentiate "relation" with different swedish labels

niklasl:
Yes we should! Suggestions?

niklasl:
I made a proposal in https://github.com/libris/definitions/pull/603/changes/569fee1c265b104120dadcbbc7130f7187c6a52a (labelled the relation notion as `"relationstyp"@sv`).

livdam:
I was going to make more or less the same suggestion

PR #604, opened 2026-06-09
Add mimimal configuration for SRU/XSearch

kwahlin:


olovy (commented):
SRU should also have ?
```
      "defaultSiteFilters": [
        { "filter": "\"rdf:type\":Instance" }
      ],
```

PR #606, opened 2026-06-11
feat(search): Default to including eplikt

olovy:
Until the condition has been refined / renamed / reformulated.


kwahlin (approved):
LGTM

PR #614, opened 2026-07-07
Add RDA term descriptions extracted from Box notes

niklasl:
This adds RDA term descriptions, extracted from a ZIP download of the box notes.

See the amended script for how it was done. It can be further modified and re-run if needed; or we can use the results as the new source and directly work more on that.

Before merging, we need to consider the new types, and how to merge these with the various existing enumerations. Then we need to reference the results as new term lists to be used for cataloguing; both as datasets in this repository, and in documentation.

### DIFF
--- a/source/rda/terms.ttl
+++ b/source/rda/terms.ttl
@@ -1,0 +1,4370 @@
+prefix : <https://id.kb.se/vocab/>
+base <https://id.kb.se/term/rda/>
+
+<aspectratiodesignation/FullScreen> a :AspectRatioDesignation ;
+  :prefLabel "Full screen"@en ;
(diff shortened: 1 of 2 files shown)


source/rda/terms.ttl
```diff
@@ -0,0 +1,4371 @@
+prefix : <https://id.kb.se/vocab/>
+base <https://id.kb.se/term/rda/>
+
```
niklasl:
```suggestion
# TODO: These are currently (2026-08) unapproved and unused. Do not link to these or include them in a dataset. We intend to incorporate these, subset by subset, as we move forward with work(/instance categorization and controlled lists for details.
```

livdam (approved):

